PR #2, opened 2019-09-06
Remove target_compatible_with restriction

omsmith:
This restriction isn't particularly valid in the long run. You could consider trying to do it when building .NET Framework targets, but given you can use mono as a runtime, that still isn't particularly correct


j3parker:
Ya, I like the idea of using Mono to run netfx stuff on not-Windows.
Thanks for removing this. It's a hold-over from before we were considering non-Windows use cases and it "made sense at the time" (it probably didn't, but oh well.)

PR #4, opened 2019-09-06
use standard framework compatability table, despite caveats

omsmith:


j3parker:
@omsmith think he knows what this is about and it's not drastic.

We probably want to match msbuild exactly rather than second guess it here. If we made this decision later it would be a breaking change, and we probably want to avoid the possibility of that more.

Issue #8, opened 2019-09-06
Add .NET Core frameworks < 2.1


j3parker:
See also #6 (.NET Framework) and #7 (.NET Standard)

This may be similar to .NET Standard, I'm not sure. .NET Framework is only available as reference assemblies (because it is usually installed system wide by MSIs, etc.) and .NET Standard is only available as reference assemblies (we think) because that's just how .NET Standard works (it is only interfaces, you can't actually run .NET Standard).

We couldn't find .NET Core reference assemblies with casual googling. At the same time, we might want to download non-reference-assemblies anyway so we can build self-contained exes (this would probably be a boolean attr on `csharp_binary`).

Some more investigation is required.

j3parker:
2.1 -> 3.0 got added in #56

j3parker:
Moved from "Generally usable" to "v1" because I think supporting the recent frameworks is enough to count as generally usable.

j3parker:
Although we didn't implement this I think it's safe to call this a "wontfix": .NET Core prior to 2.1 is probably not a high-demand thing. We support 2.1+ forward (at the time of writing, including the latest 5.0).

Issue #7, opened 2019-09-06
Add .NET Standard frameworks


j3parker:
See also #6 (.NET Framework) and #TODO (.NET Core)

The [NETStandard.Library](https://www.nuget.org/packages/NETStandard.Library) package has reference assemblies for .NET Standard. The layout doesn't match .NET Framework, so this will be slightly different. It looks like that is a meta-package and we will need to download it's dependencies.

j3parker:
Mostly blocked on #19

j3parker:
#44 added support for `netstandard2.0`. It was easy because things get a lot simpler at that point. We should probably still enable versions before 2.0.

2.1 is not yet published in the `NetStandard.Library` package.

j3parker:
Moved from "Generally usable" to "v1" because I think supporting the recent frameworks is enough to count as generally usable.

PR #23, opened 2019-09-07
list netstandard and netcore nuget packages

omsmith:
TODO:
- [x] alphabetize the list
- [ ] maybe different names?

I didn't list `Microsoft.NETCore.Platforms` because I wasn't sure if anything actually depended on it for realsies (easily addable once we try to actually use these?). I also didn't list `Microsoft.NETCore.Targets`, since I assume it's just msbuild targets.

### DIFF
--- a/csharp/private/repositories.bzl
+++ b/csharp/private/repositories.bzl
@@ -98,12 +98,42 @@ def csharp_repositories():
         sha256 = "912f6eed993e77c83cb8b92db72d50df7d06e4c4a02486474eae460728291989",
     )
 
-	nuget_package(
-		name = "netstandard2plus"
-		package = "NetStandard.Library"
-		version = "2.0.3"
-		sha256 = "3eb87644f79bcffb3c0331dbdac3c7837265f2cdf58a7bfd93e431776f77c9ba"
-	)
+    # .NET Core
+    nuget_package(
+        name = "Microsoft.NETCore.App"
+        package = "Microsoft.NETCore.App"
+        version = "2.2.6"
+        sha256 = "9b0ede75a9cc8311d503172ffa0177dc8b00bbfd2f3586a6b7d6152c0005f5ad"
+    )
+
+    nuget_package(
+        name = "Microsoft.NETCore.DotNetAppHost"
+        package = "Microsoft.NETCore.DotNetAppHost"
+        version = "2.2.6"
+        sha256 = "a91c0aa20a2665f4475cc66b92b9a4f1a908d1f1d711f0163f4e9ca3366a97e8"
+    )
+
+    nuget_package(
+        name = "Microsoft.NETCore.DotNetHostPolicy"
+        package = "Microsoft.NETCore.DotNetHostPolicy"
+        version = "2.2.6"
+        sha256 = "2e0eb87a29502de80a3d06a74b475158d18d5f87e2e0bbe95b9e68feefbd820c"
+    )
+
+    nuget_package(
+        name = "Microsoft.NETCore.DotNetHostResolver"
+        package = "Microsoft.NETCore.DotNetHostResolver"
+        version = "2.2.6"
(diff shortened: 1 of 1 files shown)


csharp/private/repositories.bzl
```diff
@@ -98,6 +98,391 @@ def csharp_repositories():
         sha256 = "912f6eed993e77c83cb8b92db72d50df7d06e4c4a02486474eae460728291989",
     )
 
+	nuget_package(
+		name = "netstandard2plus"
+		package = "NetStandard.Library"
```
omsmith: This has no dependencies for netstandard2.0, including everything inline in `build/netstandard2.0/ref`. It also contains this targets file:

```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <NETStandardLibraryPackageVersion>2.0.3</NETStandardLibraryPackageVersion>
  </PropertyGroup>

  <!-- Only add references if we're actually targeting .NETStandard.
       If the project is targeting some other TFM that is compatible with NETStandard we expect
       that framework to provide all references for NETStandard, mscorlib, System.* in their own 
       targeting pack / SDK.
       Choose/When is required to avoid an issue where the old desktop project system in VS  
       will evaluate this item ignoring the condition in order to populate the references. --> 
  <Choose>
    <When Condition="'$(TargetFrameworkIdentifier)' == '.NETStandard'">
      <ItemGroup>
        <Reference Include="$(MSBuildThisFileDirectory)ref\netstandard.dll">
          <!-- Private = false to make these reference only --> 
          <Private>false</Private>
          <!-- hide these from Assemblies view in Solution Explorer, they will be shown under packages --> 
          <Visible>false</Visible>
          <NuGetPackageId>NETStandard.Library</NuGetPackageId>
          <NuGetPackageVersion>$(NETStandardLibraryPackageVersion)</NuGetPackageVersion>    
        </Reference>
```

csharp/private/repositories.bzl
```diff
@@ -98,6 +98,391 @@ def csharp_repositories():
...
+		version = "2.0.3"
+		sha256 = "3eb87644f79bcffb3c0331dbdac3c7837265f2cdf58a7bfd93e431776f77c9ba"
+	)
+
+    nuget_package(
+        name = "System.Collections"
+        package = "System.Collections"
```
omsmith: The rest of these are what `NETStandard.Library` depends on (directly or transitively). They each have varying targeted frameworks available in `ref/`.

omsmith: Some of these may not fully apply to netstandard, and only be here for netcore, but I didn't necessarily get every package netcore needs.

j3parker:
Relevant to #7 (and #8 I guess).
Cool. It's going to be interesting getting these wired into the `@net` targets :P

j3parker:
total nit: maybe some private helper macros to separate the different packages a bit

j3parker:
So I'm thinking we check in a bunch of `BUILD`s relevant to each package? Check them in so we can generate `deps` correctly.

I still think it'd be good to wire these into `@net` targets eventually to make multi-targeting easy (e.g. at D2L we'll want to multi-target .NET framework and standard for some DLLs as we do that conversion). What do you folks think?

omsmith:
> total nit: maybe some private helper macros to separate the different packages a bit

Can you expand on that one a bit?

j3parker:
Oh sorry, yeah that's unclear. I meant a helper function like `_netstandard_repositories`... but then there is overlap with .NET Core and maybe `csharp_repositories` should just be a big alphabetical list of packages anyway. Ignore that comment :P

(btw: I think the idea behind these `_repositories` macros is that it allows consumers of these rules to not call them, and instead provide their own definition of what the `@System.Collections` is, etc., if they want to go wild).

j3parker:
Cool! Keeping these up-to-date will be an interesting challenge :)

omsmith:
> Cool! Keeping these up-to-date will be an interesting challenge :)

Yeah, can probably build a tool to walk the dependency tree of a package for a set of tfms and output the rules.

PR #26, opened 2019-09-08
Don't allow netstandard framworks for executables

### LINKED ISSUE #5: csharp_binary/csharp_nunit_test should not allow .NET Standard in their target_frameworks
This shouldn't work:

```python
csharp_binary(
  name = "MyExe",
  srcs = [/* ... */],
  target_frameworks = ["netstandard2.0"],
)
````

because .NET Standard can't be used to produce executable things. It should be possible to _reference_ .NET standard assemblies from these rules, however.

This could be implemented by adding a check inside the rules. We might want a dict where the keys are .NET Standard TFMs to support this.

j3parker:
This just makes sure we have a good error message if the user tries it.
Fixing this brought to light that we don't actually support .NET
Standard tests because we are producing an exe for Bazel to run. We
should still do that, but without modifying the users assembly (our shim
can link the tests in at runtime).

Fixes #5.

### DIFF
--- a/csharp/private/common.bzl
+++ b/csharp/private/common.bzl
@@ -16,6 +16,9 @@ def is_debug(ctx):
     # equivalent to dbg right now but might want to support this in the future.
     return ctx.var["COMPILATION_MODE"] != "opt"
 
+def is_standard_framework(tfm):
+    return tfm.startswith("netstandard")
+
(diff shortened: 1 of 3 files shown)


j3parker:
I updated the error message because it didn't really make sense, thanks for the catch.

PR #28, opened 2019-09-09
Configure NuGet files by default

### LINKED ISSUE #22: Create a good default BUILD file for NuGet packages
We should do what we did for things like [`nunitframework.BUILD`](https://github.com/Brightspace/rules_csharp/blob/master/csharp/private/nunit/nunitlite.BUILD) automatically, via a macro that uses globs.

The downside to this is that the `import_library` rules won't have `deps` set up (it's not possible for us to do this within Bazel). It's on the end user to reference all the things they need (and the transitive deps for nuget packages). We still have `build_file`/`build_file_content` as an escape hatch to do more sophisticated things, and maybe offer a tool to generate these from `packages.config`/the centralized packages stuff.

@mthjones mentioned that NuGet packages can include a ref folder tree, so we should use that. He also mentioned there are other weird folders (like runtimes, native) etc. and we should consider having our auto-generated things fail during loading when those are there (but maybe that will be annoying, so give this some thought).

j3parker:
This has caveats (we don't set up deps correctly because we can't) but
should be a reasonable default thing to do.

It might make sense to fail() if there are some unsupported things in
this package, like native/. We might also want to export the content
files.

Fixes #22


csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
+load("//csharp/private:providers.bzl", "CSharpAssembly")
+load("//csharp/private:rules/imports.bzl", "import_library", "import_multiframework_library")
+
+def _import_dll(dll, has_pdb, imports):
+    path = dll.split("/")
+
+    tfm = path[1]
+
+    # Ignore frameworks we don't support (like net35)
```
j3parker: I don't know why we don't support net35, lol. It's provided by the NUnit package which is partly why it got handled here.

omsmith: Microsoft doesn't provide a ReferenceAssemblies package for net35.

j3parker: Ah, that would explain it.

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
...
+
+       We are limited by the fact that Bazel does not allow the analysis phase to
+       read the contents of source files, e.g. to correctly configure deps. For
+       more advanced usages a BUILD file will need to be generated outside of
+       Bazel.
+
+       This has to be public so that packages can call it but you probably
+       shouldn't use it directly.
+    """
+    dlls = native.glob(["**/*.dll"])
```
omsmith: Globs are relative to the caller?

omsmith: Probably don't need to care right away, but what does a failure look like if one of these is say, a directory?

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
...
+        _import_dll(dll, has_pdb, imports)
+
+    # Output import_multiframework_library rules
+    for (name, tfms) in imports.items():
+        import_multiframework_library(
+            name = name,
+            visibility = ["//visibility:public"],
+
+            # I don't know how to do this better.
+            netstandard = tfms.get("netstandard"),
```
omsmith: `dict.get` defaults to `None?`

csharp/private/macros/setup_basic_nuget_package.bzl
omsmith: Will this pick up refdlls? Should we only be looking in `lib`?

j3parker: Oh shoot I had this as `lib/*/*.dll` at one point but was messing around. Will fix.

csharp/private/macros/setup_basic_nuget_package.bzl
j3parker: Apparently. I'm not a Python expert, I found this via Google ;)

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
...
+    for dll in dlls:
+        _import_dll(dll, has_pdb, imports)
+
+    # Output import_multiframework_library rules
+    for (name, tfms) in imports.items():
+        import_multiframework_library(
+            name = name,
+            visibility = ["//visibility:public"],
+
+            # I don't know how to do this better.
```
jrbeverly: I believe this can be replacing using [keyword arguments](https://docs.python.org/3/tutorial/controlflow.html#keyword-arguments). You can see the bazel usage for [miniature](https://docs.bazel.build/versions/master/skylark/tutorial-creating-a-macro.html) has something based on this.

```python
    for tfm in tfms:
        kwargs[tfm.replace(".", "_")] = tfms[tfm]
```

Then `**kwargs` can be used like so:

```python
import_multiframework_library(
    name = name,
    visibility = ["//visibility:public"],
    **kwargs
)
```

Our internal implementation of `copy_files` used it while it was a macro (it no longer does as it doesn't have windows/linux conditionals anymore)

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
...
+    if dll in has_pdb:
+        import_library(
+            name = target_name,
+            target_framework = tfm,
+            dll = dll,
+            pdb = dll[:-3] + "pdb",
+        )
+    else:
+        import_library(
+            name = "%s-%s" % (lib_name, tfm),
```
jrbeverly: Is this name intended to be different, or can it also be set to `target_name`?

```suggestion
            name = target_name,
```

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -0,0 +1,94 @@
...
+
+       We are limited by the fact that Bazel does not allow the analysis phase to
+       read the contents of source files, e.g. to correctly configure deps. For
+       more advanced usages a BUILD file will need to be generated outside of
+       Bazel.
+
+       This has to be public so that packages can call it but you probably
+       shouldn't use it directly.
+    """
+    dlls = native.glob(["lib/*/*.dll"])
```
jrbeverly: For nuget packages, how are envisioning the handling of [runtime + other](https://github.com/dotnet/corefx/tree/master/pkg/Microsoft.NETCore.Platforms#rid-qualified-assets-are-preferred)  assets?

Will this be including some sort of priority analysis similar to how [FrameworkCompatibility ](https://github.com/Brightspace/rules_csharp/blob/master/csharp/private/providers.bzl#L88) exists?

csharp/private/macros/setup_basic_nuget_package.bzl
j3parker: Cool, thanks! Let's try it in a follow-up.

csharp/private/macros/setup_basic_nuget_package.bzl
j3parker: Thanks, left-over from refactoring.

csharp/private/macros/setup_basic_nuget_package.bzl
j3parker: For runtimes my gut feel is that we'd just change which DLL we put in place for the `dll` attr of `import_library`. That might be tricky because I'm not sure if the loading phase knows about your target platform yet. Alternatively I think we'd only need to make this a feature of `import_library` but I'm not confident.

For now I don't think we need to support it but it would be good to error out if that folder exists. None of the packages that D2L uses in the mono-repo have runtime-specific variants.

PR #36, opened 2019-09-11
Add some basic CI

### LINKED ISSUE #18: Set up tests and pre-submit CI
We could start with buildifier lint.

Let's use GitHub actions. We can define some Bazel-related actions in another repo and request to upstream this into bazelbuild.

j3parker:
Fixes #18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,12 +5,18 @@ jobs:
   build:
     name: Build examples
     runs-on: ubuntu-latest
-    container: gcr.io/cloud-marketplace-containers/google/bazel:0.29.1
+    container:
+      image: gcr.io/cloud-marketplace-containers/google/bazel:0.29.1
     steps:
       - name: Clone
         uses: actions/checkout@v1
-      - name: Build all targets
-        run: bazel build //examples:*
       - name: Dump the action graph
         run: bazel aquery 'deps(//examples:*)'
+      - name: Build all targets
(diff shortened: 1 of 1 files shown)


j3parker:
I need to mount the source code from the checkout job into the Bazel container. I haven't used `container:` before so I need to read the docs to refresh my memory on how the paths work in actions.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: CI
+on: pull_request
+
+jobs:
+  checkout:
+    name: Clone the repo
```
j3parker: I did this in parallel in the hopes that GitHub could be downloading the Bazel container in parallel but in retrospect that obviously isn't how actions jobs work I'll merge these jobs for simplicity.

j3parker:
```
Couldn't find a valid ICU package installed on the system. Set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support.
```

Neat!

j3parker:
Ah, setting that environment variable in the container doesn't work because the action runs in a "sandbox" without those envvars, right.

j3parker:
https://bazel.build/designs/2016/06/21/environment.html

j3parker:
Something is better than nothing. I can set this as a required check.
I can work on `bazel build` in a draft PR separately.

j3parker:
😮 get those away from CI, you'll break it!!!!

PR #33, opened 2019-09-11
set up an @net repository

omsmith:
This doesn't work as is (`no such package '@net472//'`)... but I think this is some sort of thing on the right path?

PR #37, opened 2019-09-12
Add Diamond example, and fix a design bug

j3parker:
Whew! I added `examples/diamond/`in excitement after #33 and when I did a build I got errors like:

```
ERROR: file 'examples/diamond/Bottom.dll' is generated by these conflicting actions:
Label: //examples/diamond:Bottom
RuleClass: csharp_library rule
Configuration: 9db7557022224156d4ba00ff45ffc062
Mnemonic: CSharpCompile
Action key: 9dc446100544dbedb8f7a06179803675, d9921dee1e6ce573097cfcfbbd35e621
Progress message: Compiling Bottom
PrimaryInput: File:[/Users/jparker/rules_csharp[source]]examples/diamond/Bottom.cs
PrimaryOutput: File:[[<execution_root>]bazel-out/darwin-fastbuild/bin]examples/diamond/Bottom.dll
Owner information: //examples/diamond:Bottom BuildConfigurationValue.Key[9db7557022224156d4ba00ff45ffc062] false
MandatoryInputs: are equal
Outputs: are equal
```

It turns out, kinda obviously, that only one action should generate a particular file. This PR also makes the outputs distinct by adding the TFM. My bad!


csharp/private/actions/assembly.bzl
```diff
@@ -52,9 +52,10 @@ def AssemblyAction(
         toolchain):
     out_ext = "dll" if target == "library" else "exe"
 
-    out = actions.declare_file("%s.%s" % (name, out_ext))
-    refout = actions.declare_file("%s.ref.%s" % (name, out_ext))
-    pdb = actions.declare_file(name + ".pdb")
+    sanitized_tfm = target_framework.replace(".", "_")
+    out = actions.declare_file("%s.%s.%s" % (name, sanitized_tfm, out_ext))
```
omsmith: Don't suppose it's possible to put them in separate directories (like is the default behaviour for msbuild/csproj)

j3parker: I don't think so. It definitely can't cross package boundaries which would make it real awkward.

j3parker: https://docs.bazel.build/versions/master/skylark/lib/actions.html#declare_file seems to indicate it could be. I'll try making the path be the TFM

j3parker:
```
$ bazel build :Top
INFO: Analyzed target //examples/diamond:Top (13 packages loaded, 69 targets configured).
INFO: Found 1 target...
Target //examples/diamond:Top up-to-date:
  bazel-bin/examples/diamond/net48.bazelout/Top.dll
  bazel-bin/examples/diamond/net48.bazelout/Top.ref.dll
  bazel-bin/examples/diamond/net48.bazelout/Top.pdb
INFO: Elapsed time: 4.276s, Critical Path: 3.87s
INFO: 6 processes: 6 darwin-sandbox.
INFO: Build completed successfully, 7 total actions
```

Thanks. That's a better outcome. I added `.bazelout` to at make it less likely to collide with a "real" directory.

.github/workflows/ci.yml
```diff
@@ -11,4 +11,4 @@ jobs:
       - name: Clone
         uses: actions/checkout@v1
       - name: Dump the action graph
-        run: bazel aquery 'deps(//examples:*)'
+        run: bazel aquery 'deps(...)'
```
j3parker: After this change, `//examples` is no longer a package. This makes it analyze "more" stuff.

Issue #19, opened 2019-09-06
Set up deps inside .NET Framework DLLs correctly


j3parker:
Our framework build files (like [`net472.BUILD`](https://github.com/Brightspace/rules_csharp/blob/master/csharp/private/frameworks/net472.BUILD)) don't have dependencies set up correctly. They were hacked up a bit to avoid telling Bazel about .NET Framework's circular dependencies.

We need to fix this. I know how to do it I just haven't had time to sit down and fix the generator.

We will break the cycles in `deps`, but create [`alias`](https://docs.bazel.build/versions/master/be/general.html#alias) targets.

For example, `@net472//:System.Xml` will depend on `@net472//:System.Data.SqlXml-somesuffixhere`, but `@net472//:System.Data.SqlXml-somesuffixhere` won't dep on `@net472//:System.Xml` (even though it does in a CLR sense). We will do `alias("System.Data.SqlXml", actual = ":System.Xml")` so that when you dep on `@net472//:System.Data.SqlXml` all the right things happen.

PR #39, opened 2019-09-13
Frameworks

### LINKED ISSUE #6: Add additional .NET Frameworks
See also #7 (.NET Standard) and #8 (.NET Core)

We currently have `@net472`. We should do the same thing for the other .NET Framework versions.

This will be largely copy/paste work. We will need to (possibly not immediately) regenerate the `net472.BUILD` file for those frameworks too.

.NET Standard and .NET Core work slightly differently than .NET Framework in this respect so I'll open up a different issue.

omsmith:
Closes #6 and #19

This skips `net40` since I was having trouble loading those assemblies (assembly already loaded). Should open another issue to track adding that.


csharp/private/frameworks/net20.BUILD
```diff
@@ -0,0 +1,668 @@
+# This file was generated by build\src\D2L.Build.BazelGenerator\FrameworkBuildGenerator\GenerateFrameworkBuildFile.cs
```
j3parker: Each of these was independently generated?

csharp/private/frameworks/net20.BUILD
```diff
@@ -0,0 +1,668 @@
...
+    target_framework = "net20",
+    refdll = "//:build/.NETFramework/v2.0/sysglobl.dll",
+    deps = [
+        "//:mscorlib",
+        "//:System.Xml",
+    ],
+    visibility = ["@net//:__pkg__"],
+)
+
+alias(
```
j3parker: @jimevans so this is how we're dealing with cycles: When you reference `//:System` you're actually referencing `//:System.Xml`. That's OK because `//:System.Xml` has a reference to `//:System-noncyclic` (the next rule below). That rule doesn't have a dep on `//:System` because Bazel (reasonably) won't allow cycles.

csharp/private/frameworks/net20.BUILD
```diff
@@ -0,0 +1,668 @@
...
+    name = "System",
+    actual = "//:System.Xml",
```
j3parker: Does this also need `visibility = ["@net//:__pkg__"],`?

csharp/private/frameworks/net20.BUILD
omsmith: Correct

csharp/private/frameworks/net20.BUILD
omsmith: Probably

csharp/private/frameworks/net472.BUILD
```diff
@@ -1,2118 +1,3158 @@
...
+# This file was generated by build\src\D2L.Build.BazelGenerator\FrameworkBuildGenerator\GenerateFrameworkBuildFile.cs
 # Don't edit it by hand! You should regenerate this if the .NET package gets updated.
 
 load("@d2l_rules_csharp//csharp:defs.bzl", "import_library")
 
 import_library(
     name = "Accessibility",
-    refdll = ":build/.NETFramework/v4.7.2/Accessibility.dll",
     target_framework = "net472",
+    refdll = "//:build/.NETFramework/v4.7.2/Accessibility.dll",
```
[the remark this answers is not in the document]
j3parker: It's equivalent here, no problem.

PR #44, opened 2019-09-14
Add netstandard2.0 to @net

j3parker:
So it turns out that all the `netstandard2.0` assemblies are type-forwarded to
`netstandard.dll`. That means our `BUILD` file for `netstandard2.0` can be
quite simple.

I verified that every DLL in `NetStandard.Library` already had a target  in
`@net` with this command:

```sh
ls $NetStandardLibrary/build/netstandard2.0/ref/*.dll \
| sed 's:\.dll$::' \
| xargs -n1 dash -c 'grep -q "name = \"$0\"" ~/rules_csharp/csharp/private/net/BUILD || echo $0'
```

Which prints out the name of missing targets, and I got no lines of output. I
generated a batch file for [buildozer](https://github.com/bazelbuild/buildtools/tree/master/buildozer) with this command:

```sh
ls $NetStandardLibrary/build/netstandard2.0/ref/*.dll \
| sed 's:.*/::' | sed 's:\.dll$::' \
| sed 's_^_set netstandard20 "@NetStandard.Library//:netstandard"|_' \
> replacements.txt
```

and I ran the output of that with `buildozer -f` in the `csharp/private/net`
folder/workspace.

Buildozer wanted to format the files first, so I let it do that but checked it
in as the first commit. The diff of the second commit is the interesting stuff.

### DIFF
--- a/csharp/private/net/BUILD
+++ b/csharp/private/net/BUILD
@@ -346,6 +346,7 @@ import_multiframework_library(
     net471 = "@net471//:Microsoft.Win32.Primitives",
     net472 = "@net472//:Microsoft.Win32.Primitives",
     net48 = "@net48//:Microsoft.Win32.Primitives",
+    netstandard2_0 = "@NetStandard.Library//:netstandard",
     visibility = ["//visibility:public"],
 )
 
@@ -363,6 +364,7 @@ import_multiframework_library(
     net471 = "@net471//:mscorlib",
     net472 = "@net472//:mscorlib",
     net48 = "@net48//:mscorlib",
+    netstandard2_0 = "@NetStandard.Library//:netstandard",
     visibility = ["//visibility:public"],
 )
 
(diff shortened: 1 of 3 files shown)


j3parker:
Hah, GitHub interprets @net as pinging that user. Sorry dude. You might want to filter out notifications for this repo :(

csharp/private/net/BUILD
```diff
@@ -2,8 +2,6 @@ load("@d2l_rules_csharp//csharp/private:rules/imports.bzl", "import_multiframewo
 
 import_multiframework_library(
     name = "Accessibility",
-    visibility = ["//visibility:public"],
-
```
j3parker: This is what I meant by the buildifier-ish changes. I didn't feel like fighting the tools so these changes are in the first commit.

j3parker:
Hm, which decision?

omsmith:
I mean the decision to just go straight to `netstandard.dll`, which differs from msbuild.

j3parker:
Oh, yeah. I think it will work out the same for the most part but it might be worth doing the "proper" thing. Doing it this way probably changes the list of loaded assemblies if you go poking around.

Issue #41, opened 2019-09-13
Add default references like System


j3parker:
We pass `/noconfig` to csc to avoid getting [it's default references](https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/csc/csc.rsp). This matches the behaviour of MSBuild. However, MSBuild itself adds some default references too (a "standard library" of sorts). We should match that behaviour.

It will probably have to be a framework-specific list. Possibly do this via a `target_group` for each framework.

omsmith:
Here's the `Reference` `NETStandard.Library` adds, for reference

```xml
[...]

PR #47, opened 2019-09-18
Arg tweaks to match MSBuild

j3parker:
A minor grab-bag.

### DIFF
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -4,7 +4,7 @@ load(
     "get_analyzer_dll",
     "use_highentropyva",
 )
-load("@d2l_rules_csharp//csharp/private:providers.bzl", "CSharpAssembly")
+load("@d2l_rules_csharp//csharp/private:providers.bzl", "CSharpAssembly", "SubsystemVersion")
 
 def _format_ref_arg(assembly):
     return "/r:" + assembly.path
@@ -74,6 +74,10 @@ def AssemblyAction(
         args.add("/highentropyva+")
     else:
         args.add("/highentropyva-")
+
+    ssv = SubsystemVersion[target_framework]
+    if ssv != None:
+        args.add("/subsystemversion:" + ssv)
 
     args.add("/warn:0")  # TODO: this stuff ought to be configurable
 
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -83,12 +83,13 @@ def AssemblyAction(
     if debug:
         args.add("/debug+")
         args.add("/optimize-")
+        args.add("/define:TRACE;DEBUG")
     else:
+        args.add("/debug-")
         args.add("/optimize+")
+        args.add("/define:TRACE;RELEASE")
 
-        # TODO: .NET core projects use debug:portable. Investigate this, maybe move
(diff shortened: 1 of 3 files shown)


csharp/private/actions/assembly.bzl
```diff
@@ -78,12 +87,13 @@ def AssemblyAction(
...
+        args.add("/define:TRACE;DEBUG")
     else:
+        args.add("/debug-")
         args.add("/optimize+")
+        args.add("/define:TRACE;RELEASE")
 
-        # TODO: .NET core projects use debug:portable. Investigate this, maybe move
-        #       some of this into the toolchain later.
-        args.add("/debug:pdbonly")
+    args.add("/debug:portable")
```
j3parker: This one is debatable. SDK-style (what we call old style) use pdbonly, but what we call modern-style use portable. Not all tools support the portable PDB format yet (?) hence the difference.

It seems most reasonable to just do the future thing. We could make it configurable if we have to. Obviously we'll need to put some more thought into this for D2L (but then again, we'll be 100% modern-style soon I hope). I can back this part of the change out for now if you want.

csharp/private/common.bzl
```diff
@@ -16,6 +16,9 @@ def is_debug(ctx):
     # equivalent to dbg right now but might want to support this in the future.
     return ctx.var["COMPILATION_MODE"] != "opt"
 
+def use_highentropyva(tfm):
+    return tfm not in ["net20", "net40"]
```
j3parker: I figure this list is really unlikely to change, which is why I didn't make it a lookup table like the rest.
But maybe Microsoft will release reference assemblies for `net35`?
If you think this should be a lookup table like the rest (so that we have to explicitly choose) I can do that.

j3parker:
I generated a csproj for every framework and dumped the debug output from MSBuild. These are the easy (relevant) differences. The rest are in the default references (see #41).

csharp/private/actions/assembly.bzl
```diff
@@ -83,12 +83,13 @@ def AssemblyAction(
     if debug:
         args.add("/debug+")
         args.add("/optimize-")
+        args.add("/define:TRACE;DEBUG")
```
omsmith: Aren't these defined in the `csproj`?

omsmith:
> Leading zeroes in the minor version don't change the version, but trailing zeroes do. For example, 6.1 and 6.01 refer to the same version, but 6.10 refers to a different version. We recommend expressing the minor version as two digits to avoid confusion.

🤣

csharp/private/actions/assembly.bzl
j3parker: In modern style csproj it’s implict. So, yes and no.

j3parker:
Wow wtf

csharp/private/actions/assembly.bzl
omsmith: Hmm okay. Fine for now then I guess.

I was sort of thinking people should just define them on their rules. Is there a way to _undefine_ them for a dotnet csproj?

j3parker: Probably via the update/remove mechanism. I support adding this to our rules too!

We talked IRL about how we'll probably want a `stdlib = False` flag on our rules once we auto-add the stdlib for similar reasons.

PR #48, opened 2019-09-19
Add import_nuget_package

### LINKED ISSUE #29: Allow checked in nupkg to be used in nuget_package
`nuget_package` is built on top of `http_archive` which allows a file to be used in place of a URL.

We should mimic this so that people can check their nupkg's in/get them locally if they desire. Maybe this could be an optional `src` attr, and use it in place of nuget.org?

j3parker:
Fixes #29


csharp/defs.bzl
```diff
@@ -21,7 +21,8 @@ load(
     _import_multiframework_library = "import_multiframework_library",
 )
 load(
-    "//csharp/private:rules/nuget.bzl",
+    "//csharp/private:macros/nuget.bzl",
```
j3parker: I moved this file into macros from the rules folder because... it only has macros. 🤷‍♂

WORKSPACE
```diff
@@ -4,8 +4,24 @@ load(
...
 
 csharp_repositories()
 
 csharp_register_toolchains()
+
+# These are some examples.
+# Should this go here? Should we have a WORKSPACE inside csharp/ and one in
+# examples/?
+
+import_nuget_package(
```
j3parker: This case wasn't mentioned in the issue but will work very well for D2L, I think.

In the transitory period with Bazel we end up downloading the packages twice (once for the Bazel and again for the non-Bazel parts of our build) which sucks. Using this we can just wire up the extracted packages that live in `packages/` for now and save time/bandwidth.

j3parker:
From CI output:

```
action 'Compiling Demo'
  Mnemonic: CSharpCompile
  Target: //examples/import_nuget_package:Demo
  Configuration: k8-fastbuild
  ActionKey: a257c0101f4a79a021320bd9989f91f4
  Inputs: [examples/import_nuget_package/Class.cs, external/ExampleNupkg/lib/netstandard2.0/SomePackage.dll, external/ExamplePackageFolder/lib/netstandard2.0/SomePackage.dll, external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll, external/netcore-runtime-linux/dotnet]
  Outputs: [bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.dll, bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.pdb, bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.ref.dll]
  Command Line: (exec external/netcore-runtime-linux/dotnet \
    external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll \
    /noconfig \
    /unsafe- \
    /checked- \
    /nostdlib+ \
    /utf8output \
    /deterministic+ \
    /filealign:512 \
    /nologo \
    /highentropyva+ \
    /subsystemversion:6.00 \
    /warn:0 \
    /target:library \
    /langversion:7.3 \
    /debug+ \
    /optimize- \
    '/define:TRACE;DEBUG' \
    /debug:portable \
    /out:bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.dll \
    /refout:bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.ref.dll 
    /pdb:bazel-out/k8-fastbuild/bin/examples/import_nuget_package/bazelout/net472/Demo.pdb \
    /r:external/ExampleNupkg/lib/netstandard2.0/SomePackage.dll \
    /r:external/ExamplePackageFolder/lib/netstandard2.0/SomePackage.dll \
    examples/import_nuget_package/Class.cs \
    /d:NETFRAMEWORK \
    /d:NET472)
```

The `/r:`'s look right to me.

WORKSPACE
```diff
@@ -4,8 +4,24 @@ load(
...
+    "import_nuget_package",
 )
 
 csharp_repositories()
 
 csharp_register_toolchains()
+
+# These are some examples.
+# Should this go here? Should we have a WORKSPACE inside csharp/ and one in
+# examples/?
```
omsmith: Maybe just in examples?

docs/UsingNuGetPackages.md
```diff
@@ -0,0 +1,28 @@
+# Using NuGet Packages
+
+There are three options for using NuGet packages:
+
+1. Use `nuget_package` download a package from the internet.
```
omsmith: Should this one also be part of `import_nuget_package`, and this name be used for producing a nupkg?

csharp/private/macros/nuget.bzl
```diff
@@ -0,0 +1,106 @@
...
+    either be a relative path in the workspace, or an absolute path. Note that
+    paths on Windows use forward slashes.
+
+    At most one of build_file or build_file_content may be provided. If neither
+    are, a default BUILD file will be used that makes a "best effort" to wire
+    up the package. See docs/UsingNuGetPacakges.md for more info.
+
+    Args:
+      name: A unique name for the package's workspace.
+      dir: A directory containing an extracted package.
```
omsmith: I think it'd be nicer to not have "custom" names here, but I guess we need to disambiguate the directory case?

docs/UsingNuGetPackages.md
omsmith: Alternatively, just give them 3 seperate names, all starting with `import_`?

j3parker: Hm, so my thought here is that I've mostly seen `import_` as a prefix (this isn't consistent... `java_import` uses it as a suffix ;) ) for "rule that brings in thing that exists on disk into Bazel". I imagine most people would use `nuget_package` and other rule sets have used the noun for their package type as the rule for downloading the package.

I don't think there is a lot of consistency though. e.g. `pip_import` actually downloads packages!

🤷‍♂

csharp/private/macros/nuget.bzl
j3parker: What do you mean by a custom name? Like `file` vs. `dir` as mutually exclusive args?
I didn't see an obvious way to do "is this path a file or folder" in Starlark but I also didn't try hard :)

WORKSPACE
j3parker: Ya probably. I'll open an issue to do that tomorrow. It's going to be easier to configure Bazel CI with one WORKSPACE to start though so I'd like to get that done first so that I can test the change in CI. I'm happy to delay this change until that's set up though.

j3parker:
@omsmith and I discussed this IRL.

We we're both looking at this from different perspectives:

* I was thinking of `nuget_package` as an abstraction around `dotnet pack`, even though it's implemented via `http_archive`
* Owen was thinking about it as a macro around `http_archive` (which it factually is right now).

I didn't have any justification for my perspective. We looked into more rule sets in the bazelbuild repo and came to the conclusion that there is no "right" way to name or organize these things; every rule-set is different.

We did agree that these three use-cases are necessary to cover. We both agreed that we probably have the flexibility to iterate on the design of these rules. I think we both agreed generally that what I have here isn't necessarily right.

I'm going to merge this now (so that I can switch our code-base to directory-based packages) but open up an issue to pin down the design before v1.

j3parker:
We also wanted to determine if our relative `file:` URLs are legit. We're relying on what appears to be undocumented behaviour, so we need to make sure that's OK. I'll reach out in the Bazel org about this and post the issue back here.

PR #49, opened 2019-09-19
Fix glob

j3parker:
It will soon be an error for a glob to return an empty set. This is
gated behind --incompatible_disallow_empty_glob currently.


j3parker:
Linux/Mac have `dotnet`, Windows has `dotnet.exe`

omsmith:
Oh, so the glob items not resolving, not the glob as a whole

PR #50, opened 2019-09-19
nuget api suggestion

omsmith:


j3parker:
Thanks. In #48 I did my best to summarize our conversation and created #52 to track finalizing the API.

PR #53, opened 2019-09-21
Support other frameworks (like .NET 3.5) in NuGet packages

j3parker:


csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -1,24 +1,47 @@
...
+    real_tfm = tfm
+    quasi_supported = False
 
-    # Ignore frameworks we don't support (like net35)
-    if tfm not in CSharpAssembly:
+    if tfm in QUASI_SUPPORTED_TFMS.keys():
+        tfm = QUASI_SUPPORTED_TFMS[tfm]
+        quasi_supported = True
+    elif tfm not in CSharpAssembly:
+        # Ignore any other frameworks (e.g. PCLs)
```
j3parker: We could support more in the same way.

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -1,24 +1,47 @@
...
+    elif quasi_supported and tfm in imports[lib_name]:
+        # Don't bother with a quasi-supported framework if we also have a DLL for
+        # its closest framework.
+        return
     else:
         imports[lib_name][tfm] = target_name
 
+    # if quasi_supported is true we might eventually overwrite
+    # imports[lib_name][tfm] which means the import_library we're about to do
+    # would go unreferenced. It's a private target though so it's not a big deal.
```
j3parker: I think this will (at least with the current list of quasi-supported frameworks) always get hit, because apparently [glob returns things in sorted order](https://docs.bazel.build/versions/master/be/functions.html#glob).
I'm fine with that, but another option would be to reverse its output and have this case never get it. Sort order may not be reliable if we want to support PCLs (e.g. "portable-net4+win8") but then again I'm not sure it'll be trivial (or necessary) for us to support those?

j3parker:
Hmm... maybe it’d be less weird to just define providers for those frameworks even though we can’t compile them(because MS hasn’t made the ref assemblies for them available)? Then there would be no extra logic. 

What do you think?

omsmith:
Define providers, or just include them in the compatability table?

j3parker:
Define providers + include them in the compatibility table.

j3parker:
so like net40 -> net403 -> net45... it would allow you to import `net403` things (just like how we can import net standard < 2.0 today) you just couldn't do `csharp_library` targetting `net403` if you want stdlib deps (which.. makes sense?)

j3parker:
I think I'm leaning more towards that solution now... because it doesn't require any extra code. Let me know what you think.

omsmith:
I don't think you actually need to define the providers though. Just adding them to the compatibility table would fall back to the correct things we _do_ support?

j3parker:
Ok, I'll give it a go. I don't like the diff in this PR so I'm going to abandon it.

PR #54, opened 2019-09-22
Fix @net for users of this repo

j3parker:
Whoops! I noticed this didn't work outside this repo.


csharp/private/rules/create_net_workspace.bzl
```diff
@@ -0,0 +1,31 @@
...
+
+local_repository doesn't work well because it takes a path that is either
+absolute (yuck) or relative to our users WORKSPACE file. Instead we use a
+custom workspace rule.
+"""
+
+def _create_net_workspace_impl(ctx):
+  # The "trick" here is that repository rules each create a workspace (named by
+  # their name attr). So when we create files with ctx it's inside that
+  # workspace.
```
j3parker: I started by doing the dumb thing and using a sacrificial NuGet package, but then felt bad and went digging in [around `repository_ctx.extract`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#extract) That and looking at how [`http_archive` works](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L59) led me to [the `workspace_and_buildfile` function](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/utils.bzl#L31) which was the basis for how this works.

WORKSPACE
```diff
@@ -1,27 +1 @@
...
-    "import_nuget_package",
-)
-
-csharp_repositories()
-
-csharp_register_toolchains()
-
-# These are some examples.
-# Should this go here? Should we have a WORKSPACE inside csharp/ and one in
-# examples/?
```
j3parker: Done!

Issue #42, opened 2019-09-13
Integrate with Bazel CI


j3parker:
[Here's the documentation](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md).

Apparently it is possible for us to request this even though the repo isn't in the bazelbuild org. This was recommended by a Bazel team member on Slack. It looks useful.

j3parker:
I've opened an issue in [bazelbuild/continuous-integration](https://github.com/bazelbuild/continuous-integration/issues/832) to add us to their CI! Exciting!

j3parker:
Woohoo, it worked! https://buildkite.com/bazel/github-dot-com-brightspace-rules-csharp/builds/1#_

I'm just going through the remaining steps to configure it in this repo.

I will probably remove our GitHub action CI once everything is good.

j3parker:
Done!

PR #55, opened 2019-09-23
Add .bazelci/presubmit.yml

j3parker:
Related to #42 

I also need to submit an issue in their repo, according to [the documentation](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md).

I've kept this pretty basic. We'd like to test on the 3 major platforms. There are some other things I'd like in the future:

- run & test targets: blocked on #9 
- buildifier linting: blocked on us formatting our code
- using remote builds (i.e. the `rbe_ubuntu1604` platform): this is probably blocked on as-of-yet-unfiled issue where we have a bit too many implicit dependencies on the exec environment (e.g. libicu) it's why we aren't doing `bazel build` inside our GitHub action (which runs in the "official" Bazel containers, which RBE may use...)

For the extra stuff that might not succeed on attempt 1 it'll be easier to add those as follow-ups, once we can trigger CI :) My goal is to just get something basic working.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,0 +1,30 @@
+---
+bazel: 0.29.1
+tasks:
+  examples-windows:
+    name: Examples on Windows
+    platform: windows
+    include_json_profile:
+    - build
+    - test
+    working_directory: examples
+    build_targets:
+    - ...
+  examples-ubuntu1804:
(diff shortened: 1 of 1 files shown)


j3parker:
I could of course make this more minimal but I'm feeling lucky ;)

Issue #61, opened 2019-09-24
Framework resolution is too .NET Standard-averse.


j3parker:
If an assembly `Y` has a `netstandard1.0` and `netstandard2.0` build, then a `netcoreapp3.0` build of `X`, which depends on `Y`, will use the `netstandard1.0` incorrectly.

This happens because of how we build the providers "bottom up" but kinda forget about how this was happening. .NET Core things always prefer .NET Core providers from their deps, even if those providers were actually providing a .NET Standard build...

Maybe the providers should track their uh, "true target framework", so that the precedence logic can compare all possible options?

See the discussion at the end of #56

j3parker:
We've puzzled out a fix for this. Nothing too crazy!

PR #64, opened 2019-09-24
newer netstandards should be preferred over older netstandards

omsmith:
Fixes: https://github.com/Brightspace/rules_csharp/issues/61

### DIFF
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -172,5 +172,5 @@ def AssemblyAction(
         deps = deps,
         transitive_refs = refs,
         transitive_runfiles = runfiles,
-        provided_by_netstandard = False
+        actual_tfm = target_framework
     )
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -172,4 +172,5 @@ def AssemblyAction(
         deps = deps,
         transitive_refs = refs,
         transitive_runfiles = runfiles,
+        provided_by_netstandard = False
     )
--- a/csharp/private/common.bzl
+++ b/csharp/private/common.bzl
@@ -77,10 +77,7 @@ def fill_in_missing_frameworks(providers):
         # nested loop isn't bad.
         # Order by providers that didn't "cross the netstandard boundary" so
         # newer netstandard will be preferred, if applicable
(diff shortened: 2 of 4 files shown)


csharp/private/common.bzl
```diff
@@ -72,12 +75,10 @@ def fill_in_missing_frameworks(providers):
 
         # There are at most 2 elements in FrameworkCompatibility[tfm], so this
         # nested loop isn't bad.
-        for compatible_tfm in FrameworkCompatibility[tfm]:
-            if compatible_tfm not in providers:
-                continue
-
+        # Order by providers that didn't "cross the netstandard boundary" so
+        # newer netstandard will be preferred, if applicable
+        for (base, compatible_tfm) in sorted([(providers[compatible_tfm], compatible_tfm) for compatible_tfm in FrameworkCompatibility[tfm] if compatible_tfm in providers], key=_get_provided_by_netstandard):
```
omsmith: No thought into line length / how to format python code. Feel free to commit fixing this (and 90)

csharp/private/actions/assembly.bzl
```diff
@@ -172,4 +172,5 @@ def AssemblyAction(
         deps = deps,
         transitive_refs = refs,
         transitive_runfiles = runfiles,
+        provided_by_netstandard = False
```
j3parker: Hm, maybe this is an awkard name because `target_framework` could be a netstandard framework...
I'm trying to think of a better name but it's awkward.

j3parker: polyfilled_by_netstandard? that's a word we haven't used so probably not good either.

omsmith: `provided_by_netstandard_but_not_netstandard_itself`
`crossed_the_netstandard_boundary`
`provided_by_alternate_framework_line`

csharp/private/common.bzl
j3parker: I've been keeping it to 80. I think buildifier can enforce this and we'll be able to run that in CI soon anyway so I'm not worried (we'll get to it)

csharp/private/actions/assembly.bzl
j3parker: `secretly_netstandard` (jk)

omsmith: Alternatively, just record the "actual tfm" and then do the logic in the sort `key=` function, thus avoiding having to name the particular situation

PR #66, opened 2019-09-25
Support net11/net30/net35/net403 from NuGet

### LINKED ISSUE #59: We need to support NuGet packages that only have net11/net35/net403.
We can't compile new DLLs for these frameworks (because MS doesn't make reference assemblies available for them yet) but we should be able to import them and use them like normal.

See #53 for an approach I abandoned because it's weird.

This is critical for D2L (if we want to avoid custom BUILD files more).

j3parker:
This is the best thing I can come up with which minimizes the amount of
hackery needed to support this without accidentally supporting compiling
using these frameworks (which we can't support because Microsoft doesn't
make reference assemblies available).

Closes #59


csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -1,19 +1,28 @@
 load("//csharp/private:providers.bzl", "CSharpAssembly")
 load("//csharp/private:rules/imports.bzl", "import_library", "import_multiframework_library")
 
+NugetOnlyFrameworks = {
```
omsmith: Should this be restricted to nuget? Should it be a feature of `import_library`?

j3parker:
This is very similar to the old PR but less hacky.

I poked around doing it the other ways and I didn't find a way that actually worked better. Some ideas seemed better but in practice they all either had their own hacks and/or implied we supported these frameworks more than we actually do.

csharp/private/macros/setup_basic_nuget_package.bzl
j3parker: I considered that but it seemed weird to support it in the `target_framework` attr for that rule only, and then you still basically need the hacks because `import_multiframework_library` doesn't support them either... you can put those hacks into `import_multiframework_library` but then it didn't really look much better.

Having the hacks inside the hacky macro that tries to do its best to make a BUILD file kinda makes sense. If you had to do this by hand you'd probably use `import_library` like how this macro is doing it.

j3parker: I think the ends justify the means: it's a hack to use `import_library` like this but it results in correct stuff. I think its still fair to say we can't really support these frameworks in general though.

csharp/private/macros/setup_basic_nuget_package.bzl
```diff
@@ -73,9 +82,9 @@ def setup_basic_nuget_package():
             netstandard1_6 = tfms.get("netstandard1.6"),
             netstandard2_0 = tfms.get("netstandard2.0"),
             netstandard2_1 = tfms.get("netstandard2.1"),
-            net20 = tfms.get("net20"),
-            net40 = tfms.get("net40"),
-            net45 = tfms.get("net45"),
+            net20 = tfms.get("net20") or tfms.get("net11"),
```
omsmith: Due to the `NugetOnlyFrameworks` translation, these `or`s will never get hit.

j3parker:
Ya this is broken

j3parker:
```
action 'Compiling MultiTarget_DepOnPackageThatHasWeirdFrameworks'
  Mnemonic: CSharpCompile
  Target: //import_nuget_package:MultiTarget_DepOnPackageThatHasWeirdFrameworks
  ...
  Command Line: (exec external/netcore-runtime-osx/dotnet \
    external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll \
    ...
    /r:external/ExamplePackageFolder/lib/net11/SomePackage.dll \
    import_nuget_package/Class.cs \
    /d:NETFRAMEWORK \
    /d:NET20)

action 'Compiling MultiTarget_DepOnPackageThatHasWeirdFrameworks'
  Mnemonic: CSharpCompile
  Target: //import_nuget_package:MultiTarget_DepOnPackageThatHasWeirdFrameworks
  ...
  Command Line: (exec external/netcore-runtime-osx/dotnet \
    external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll \
    ...
    /r:external/ExamplePackageFolder/lib/net30/SomePackage.dll \
    import_nuget_package/Class.cs \
    /d:NETFRAMEWORK \
    /d:NET40)

action 'Compiling MultiTarget_DepOnPackageThatHasWeirdFrameworks'
  Mnemonic: CSharpCompile
  Target: //import_nuget_package:MultiTarget_DepOnPackageThatHasWeirdFrameworks
  ...
  Command Line: (exec external/netcore-runtime-osx/dotnet \
    external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll \
    ...
    /r:external/ExamplePackageFolder/lib/net45/SomePackage.dll \
    import_nuget_package/Class.cs \
    /d:NETFRAMEWORK \
    /d:NET48)
```

omsmith:
Could you add `net403` to SomePackage, to demonstrate we still use the `net45`?

j3parker:
Yep!

```
action 'Compiling MultiTarget_DepOnPackageThatHasWeirdFrameworks'
  Mnemonic: CSharpCompile
  Target: //import_nuget_package:MultiTarget_DepOnPackageThatHasWeirdFrameworks
  ...
  Command Line: (exec external/netcore-runtime-osx/dotnet \
    external/csharp-build-tools/tasks/netcoreapp2.1/bincore/csc.dll \
    ...
    /r:external/ExamplePackageFolder/lib/net45/SomePackage.dll \
    import_nuget_package/Class.cs \
    /d:NETFRAMEWORK \
    /d:NET48)

File: csharp/private/macros/setup_basic_nuget_package.bzl (hunk 2 above)
Comment (omsmith) [reply csharp/private/macros/setup_basic_nuget_package.bzl]: I still think these `or`s will never be hit. `imports` isn't touched until `net11` has been mapped to `net20`

File: csharp/private/macros/setup_basic_nuget_package.bzl [hunk 3]
```diff
@@ -1,22 +1,34 @@
...
+        nuget_only_framework = True
+    elif tfm not in CSharpAssembly:
+        # Ignore other frameworks
+        return
+
     if lib_name not in imports:
         imports[lib_name] = {tfm: target_name}
-    else:
+    elif tfm not in imports[lib_name] or not nuget_only_framework:
+        # We don't overwrite for nuget-only frameworks
```
omsmith: Somewhat of a nit: We should theoretically prefer the `net35` over the `net30`. Not sure if you recognize that isn't happening, but decided this was close enough, or if you just didn't notice.

I think the behaviour as is is fine though, and an issue can be spun off.

csharp/private/macros/setup_basic_nuget_package.bzl
omsmith: Alternative phrasing: These `or`s are all unnecessary.

csharp/private/macros/setup_basic_nuget_package.bzl
omsmith: Probably fixable by iterating dlls in reverse order?

j3parker: Maybe. Let's talk about this tomorrow :)

j3parker:
Abandoned in favour of #88

Issue #15, opened 2019-09-06
Make the default langversion configurable


j3parker:
Currently we default to C# 7.3 (because that's what D2L uses). I think this should probably be configurable at the workspace level.

j3parker:
To test:

If we make a csproj with `<TargetFrameworks>net20;net48</TargetFrameworks>`, what does MSBuild do for the langauge version?

If it chooses "the latest supported language version for your framework" by default (we have some suspicions that this might be the case) does it mean you need `#ifdef`s to use newer language features?

If you set an explicit value in your csproj does it apply for all frameworks or is there any sort of ceiling logic going on? (Seems unlikely).

omsmith:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

omsmith:
Target framework | version | C# language version default
-- | -- | --
.NET Core | 3.x | C# 8.0
.NET Core | 2.x | C# 7.3
.NET Standard | all | C# 7.3
.NET Framework | all | C# 7.3

omsmith:
I'm not sure I believe that netstandard2.1 defaults to 7.3. Did they not add all the fancy features to 2.1?

PR #67, opened 2019-09-25
apply default langversion by framework

omsmith:
Closes: https://github.com/Brightspace/rules_csharp/issues/15

### DIFF
--- a/csharp/private/providers.bzl
+++ b/csharp/private/providers.bzl
@@ -154,6 +154,37 @@ SubsystemVersion = {
     "netcoreapp3.0": None,
 }
 
+DefaultLangVersion = {
+    "netstandard": "7.3",
+    "netstandard1.0": "7.3",
+    "netstandard1.1": "7.3",
+    "netstandard1.2": "7.3",
+    "netstandard1.3": "7.3",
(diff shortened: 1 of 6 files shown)


csharp/private/providers.bzl
```diff
@@ -154,6 +154,37 @@ SubsystemVersion = {
...
+    "net451": "7.3",
+    "net452": "7.3",
+    "net46": "7.3",
+    "net461": "7.3",
+    "net462": "7.3",
+    "net47": "7.3",
+    "net471": "7.3",
+    "net472": "7.3",
+    "net48": "7.3",
+    "netcoreapp1.0": "7.3",
```
omsmith: It's not in their table, but I assume

csharp/private/providers.bzl
```diff
@@ -154,6 +154,37 @@ SubsystemVersion = {
     "netcoreapp3.0": None,
 }
 
+DefaultLangVersion = {
```
omsmith: Seemed easier than doing anything fancy

j3parker: I like it simple.

PR #70, opened 2019-09-25
Add a stdlib attribute to the compilation rules

j3parker:
This is another incremental step towards #41.

The default value will be @net//:StandardLibrary eventually, and the
idea is that if you don't want this you can set the attr to be None
explicitly.

### DIFF
--- a/csharp/private/rules/binary.bzl
+++ b/csharp/private/rules/binary.bzl
@@ -10,7 +10,7 @@ load(
 def _binary_impl(ctx):
     providers = {}
 
-    stdlib = [ctx.attr.stdlib] if ctx.attr.stdlib else []
+    stdlib = [ctx.attr._stdlib] if ctx.attr.stdlib else []
 
     for tfm in ctx.attr.target_frameworks:
         if is_standard_framework(tfm):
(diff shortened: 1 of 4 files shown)


csharp/private/rules/binary.bzl
```diff
@@ -10,6 +10,8 @@ load(
 def _binary_impl(ctx):
     providers = {}
 
+    stdlib = [ctx.attr.stdlib] if ctx.attr.stdlib else []
```
j3parker: Soon it'd be reasonable to de-dupe the code in these rules... I've been putting it off until we support more stuff because I'm not sure exactly what the API should look like.

`AssemblyAction` by contrast I've been using as an interface for csc, so it's API (very) roughly matches it.

examples/diamond/BUILD
```diff
@@ -29,5 +29,10 @@ csharp_library(
 csharp_library(
     name = "Bottom",
     srcs = ["Bottom.cs"],
-    target_frameworks= ["net48", "net40", "netstandard1.4"],
+    target_frameworks= [
+        "net48",
+        "net40",
+        # TODO: change this back to "netstandard1.4" when we support it
```
j3parker: The new reference to `@net//:mscorlib` broke this example, naturally.

j3parker: Ah, I realized two things:

* I should add a test for not including the stdlib
* For now it'd be reasonable to make diamond be that test and keep it reference netstandard1.4 (which is what it does in the docs).

Doing that I discovered that `stdlib = None` wouldn't actually work. So `stdlib = False` is how you control it, and there is a private default attr `_stdlib` with the target.

If you want to "provide your own standard library" you'd add it to `deps` like normal.

omsmith:
So I think stdlib is probably close enough to correct. 

The [CLI specification](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf) does use the terminology _Standard Libraries_ (IV.5) though, of which the Framework Class Library and CoreFX are apparent implementations.

I'm not sure what these lists end up being strictly mapped to that though...? Can we find other (ideally msft) things that would call these "c#'s stdlib"? I'm just thinking of calling them _Default References_, but maybe that's too cautious.

j3parker:
`StandardReferences`?

PR #74, opened 2019-09-27
Add 'out' attribute to assembly-building commands

### LINKED ISSUE #73: Decouple output assembly name from target name
Currently, built assemblies have the same name as the name of the target that builds them. In other words,

    csharp_library(
      name = "foo",
      srcs = glob(["*.cs",
      ]),
      target_frameworks = [
        "net45",
      ],
      deps = [
        "@net://System",
      ],
    )

will always yield a built assembly of `foo.dll`. If I want the actual assembly to be named `bar.dll`, I need to do some hackery to be able to copy the file to a new name.

Proposed solution is to add an `out` attribute on the assembly-building rules, such that calling `bazel build //project:foo` will yield `bar.dll` in the output directory, if the `foo` target is defined as:

    csharp_library(
      name = "foo",
      srcs = glob(["*.cs",
      ]),
      target_frameworks = [
        "net45",
      ],
      deps = [
        "@net://System",
      ],
      out = "bar",
    )

jimevans:
This decouples the name of the built assembly from the name of the
Bazel target. If not specified, the name of the target is used.

Fixes #73.

### DIFF
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -58,7 +58,7 @@ def AssemblyAction(
         target_framework,
         toolchain):
 
-    out_file_name = name if out == None else out
+    out_file_name = name if out == "" else out
     out_dir = "bazelout/" + target_framework
     out_ext = "dll" if target == "library" else "exe"
     out_file = actions.declare_file("%s/%s.%s" % (out_dir, out_file_name, out_ext))
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -53,15 +53,17 @@ def AssemblyAction(
         langversion,
         resources,
         srcs,
+        out,
         target,
         target_framework,
         toolchain):
 
+    out_file_name = name if out == None else out
     out_dir = "bazelout/" + target_framework
     out_ext = "dll" if target == "library" else "exe"
-    out = actions.declare_file("%s/%s.%s" % (out_dir, name, out_ext))
-    refout = actions.declare_file("%s/%s.ref.%s" % (out_dir, name, out_ext))
-    pdb = actions.declare_file("%s/%s.pdb" % (out_dir, name))
+    out_file = actions.declare_file("%s/%s.%s" % (out_dir, out_file_name, out_ext))
+    refout = actions.declare_file("%s/%s.ref.%s" % (out_dir, out_file_name, out_ext))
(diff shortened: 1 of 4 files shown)


csharp/private/actions/assembly.bzl
```diff
@@ -53,15 +53,17 @@ def AssemblyAction(
         langversion,
         resources,
         srcs,
+        out,
         target,
         target_framework,
         toolchain):
 
+    out_file_name = name if out == None else out
```
j3parker: nit: I think `out or name` name would work/be idiomatic

omsmith:
Hmm, I was hoping to find a similar feature on `java_library` or `cc_library`, but didn't see one :(.

Will bazel yell if multiple targets specify the same `out` (tried to create the same output)?

j3parker:
> Will bazel yell if multiple targets specify the same out (try to create the same outputs)?

Yeah

csharp/private/actions/assembly.bzl
jimevans: Have moved from `None` to `""`, and I'm not sure how idiomatic it would be to use that construct for an empty string.

csharp/private/rules/nunit_test.bzl
```diff
@@ -70,6 +71,10 @@ csharp_nunit_test = rule(
             doc = "A list of files to embed in the DLL as resources.",
             allow_files = True,
         ),
+        "out": attr.string(
+            doc = "File name, without extension, of the built assembly.",
+            default = "",
```
omsmith: Not: This is the default default, so needn't be specified

jimevans: And yet, without it, CI fails, saying that there's no `out` attribute specified. What's the right answer for that, then?

omsmith: `out_file_name = name if out == None else out`

That's just because `out` was not `None`, so `out_file_name` became `""`. `out = out_file_name` was then invalid on `CSharpAssembly`.

Should be good now, with out without these explicit defaults.

jimevans: Ah, of course. I got in a hurry and didn't think it all the way through. Will fixup and revert the `default=""` on the rules, since they're unneeded.

jimevans: Fixed in subsequent commit

PR #76, opened 2019-09-27
Adding strong-naming via /keyfile: compiler switch

jimevans:
Allows the user to build a strong-named assembly by specifying
a label (can be a single file) to a key file.

### DIFF
--- a/csharp/private/rules/nunit_test.bzl
+++ b/csharp/private/rules/nunit_test.bzl
@@ -25,6 +25,7 @@ def _nunit_test_impl(ctx):
             debug = is_debug(ctx),
             defines = ctx.attr.defines,
             deps = ctx.attr.deps + extra_deps + stdrefs,
+            keyfile = ctx.file.keyfile,
             langversion = ctx.attr.langversion,
             resources = ctx.files.resources,
             srcs = ctx.files.srcs + [ctx.file._nunit_shim],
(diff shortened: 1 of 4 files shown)


csharp/private/rules/nunit_test.bzl
```diff
@@ -64,6 +65,10 @@ csharp_nunit_test = rule(
             doc = "A list of analyzer references.",
             providers = AnyTargetFramework,
         ),
+        "keyfile": attr.label(
```
omsmith: I can't imagine strong-naming test projects, despite it being possible. Is this something you actually do, or just being consistent?

jimevans: Mostly just being consistent here. I can't imagine doing it either, but there might be some cases where the user would want to do it.

j3parker: MSBuild allows it too ;)

omsmith: Is NUnit and NUnit Lite strong-named?

j3parker: Good question. Maybe? Random commit: https://github.com/nunit/nunit/commit/ebf531ad5bce76f3358755b4c4a304c55a2b7363

jimevans: I'm not sure. Looking at the binaries with ildasm, they don't appear to be, though I'm not certain it matters.

j3parker: Looks like it

image

j3parker: (Er, I'm assuming that PublicKey stuff is about signing... I don't actually know how this stuff works :) )

omsmith: I'm thinking the strong named test assembly won't be able to reference / load the not-strongly-named NUnit?

I may not understand strong naming though.

omsmith: > I'm assuming that PublicKey stuff is about signing

Yeah, I'd think so

PR #78, opened 2019-09-28
nuget_package: add nuget_sources arg

omsmith:


csharp/private/macros/nuget.bzl
```diff
@@ -84,15 +85,14 @@ def nuget_package(
...
       sha256: The SHA256 of the package.
       build_file: The path to a BUILD file to use for the package.
       build_file_content: A string containing the contents of a BUILD file.
     """
 
-    urls = [s + "/" + package + "/" + version for s in [
-        # TODO: allow this to be configured
-        "https://www.nuget.org/api/v2/package",
-    ]]
+    nuget_sources = nuget_sources or ["https://www.nuget.org/api/v2/package"]
```
j3parker: nice, I like that you can opt out of nuget.org

PR #80, opened 2019-09-28
add example/test that exercises every framework

omsmith:


j3parker:
This will pass analysis but probably fail `bazel build ...` because I think C# complains about not having a `.cs` file?

omsmith:
Yeah, just `warning CS2008: No source files specified.`

PR #81, opened 2019-09-28
add strong-named example

omsmith:
This isn't building successfully, but opening it because it _should_ be working?

`error CS7027: Error signing output with public key from file 'strong_named/example.snk' -- File not found.`

`csc.dll /noconfig /unsafe- /checked- /nostdlib+ /utf8output /deterministic+ /filealign:512 /nologo /highentropyva+ /subsystemversion:6.00 /warn:0 /target:library /langversion:7.3 /debug+ /optimize- '/define:TRACE;DEBUG' /debug:portable /out:bazel-out/k8-fastbuild/bin/strong_named/bazelout/net472/strong_lib.dll /refout:bazel-out/k8-fastbuild/bin/strong_named/bazelout/net472/strong_lib.ref.dll /pdb:bazel-out/k8-fastbuild/bin/strong_named/bazelout/net472/strong_lib.pdb /r:external/net472/build/.NETFramework/v4.7.2/mscorlib.dll /d:NETFRAMEWORK /d:NET472 /keyfile:strong_named/example.snk`


omsmith:
@jimevans Did you have this working locally?

jimevans:
I did, yes. I did have the .snk file exposed via an `exports_files` call, but I’m not sure that makes a difference.

csharp/private/actions/assembly.bzl
```diff
@@ -155,14 +155,16 @@ def AssemblyAction(
     args.set_param_file_format("multiline")
     args.use_param_file("@%s")
 
+    direct_inputs = srcs + resources + analyzer_assemblies + additionalfiles + [toolchain.compiler]
+    direct_inputs += [keyfile] if keyfile else []
```
j3parker: @jimevans so the problem was that it works on Windows because no sandbox, but on other platforms it failed. So we need to tell the action that the key is an input to the compiler and it works everywhere.

Issue #51, opened 2019-09-21
Add a generic csharp_test built on top of vstest


j3parker:
Right now we have `csharp_nunit_test`. There is some common infrastructure these days for test frameworks. [dotnet vstest](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-vstest?tabs=netcore21) is how most people experience it, and I believe the underlying stuff is avaialble in the [Microsoft.TestPlatform](https://www.nuget.org/packages/Microsoft.TestPlatform/) package.

Presumably we could build a generic `csharp_test` rule on top of this. `csharp_nunit_test` could become a light macro on top of that (adding a reference to nunit.framework) or go away. In either case we'd stop using the NUnitLite runner in favour of vstest.

PR #83, opened 2019-10-03
WIP: make bazel run/test run .NET Core things with dotnet.exe

### LINKED ISSUE #71: bazel run doesn't work for .NET Core on Linux and MacOS
When looking into moving my solution for #9 into this repo after #56 I noticed that `bazel run` has a few more issues:

```
/bin/bash: /private/var/tmp/_bazel_jparker/15a4c5d8b43d43bc514ff6a3dd7ab929/execroot/d2l_rules_csharp/bazel-out/darwin-fastbuild/bin/csharp/private/tools/bazel-assembly-runner/bazelout/netcoreapp3.0/bazel-assembly-runner.exe:
cannot execute binary file
```

What's happening here is `bazel-assembly-runner.exe` (the output of a `csharp_binary` rule) is actually an IL file (`dotnet build` would have given it a `.dll` extension on not-Windows). To run it, you need to pass it to the `dotnet` executable (the runtime). There is a way to create self-contained exes on Mac/Linux but it's not clear to me that we'd want to do this for every `csharp_binary` (but we'll probably want a rule to do this specifically, I'll open an issue).

If I try to execute that file with `dotnet` on Mac I get this error:

```
$ dotnet /private/var/tmp/_bazel_jparker/15a4c5d8b43d43bc514ff6a3dd7ab929/execroot/d2l_rules_csharp/bazel-out/darwin-fastbuild/bin/csharp/private/tools/bazel-assembly-runner/bazelout/netcoreapp3.0/bazel-assembly-runner.exe

A fatal error was encountered. The library 'libhostpolicy.dylib' required to
execute the application was not found in
'/private/var/tmp/_bazel_jparker/15a4c5d8b43d43bc514ff6a3dd7ab929/execroot/d2l_rules_csharp/bazel-out/darwin-fastbuild/bin/csharp/private/tools/bazel-assembly-runner/bazelout/netcoreapp3.0/'.
Failed to run as a self-contained app. If this should be a framework-dependent
app, add the /private/var/tmp/_bazel_jparker/15a4c5d8b43d43bc514ff6a3dd7ab929/execroot/d2l_rules_csharp/bazel-out/darwin-fastbuild/bin/csharp/private/tools/bazel-assembly-runner/bazelout/netcoreapp3.0/bazel-assembly-runner.runtimeconfig.json
file specifying the appropriate framework.
```

The `libhostpolicy.dylib` file exists within the runtime package that we've downloaded, for what it's worth. The way we've been going with these rules we'd be using the .NET Core 3.0 runtime for `bazel run` which I think is good. Need to investigate the things mentioned in this error message more.

j3parker:
Fixes #71 

This doesn't work yet, just wanted to show what I'm thinking.

The complexity of this vs. the goal is kind of comical to me.

### DIFF
--- a/csharp/private/tools/assembly-runner/lib.bzl
+++ b/csharp/private/tools/assembly-runner/lib.bzl
@@ -1,0 +1,147 @@
+load(
+    "@d2l_rules_csharp//csharp/private:providers.bzl",
+    "AnyTargetFramework",
+    "CSharpAssembly",
+)
+
+def _write_wrapper_cc_impl(ctx):
+    provider = CSharpAssembly[ctx.attr.target_framework]
+    exe = ctx.attr.exe[provider]
+
+    toolchain = ctx.toolchains["@d2l_rules_csharp//csharp/private:toolchain_type"]
+
+    ctx.actions.expand_template(
+        template = ctx.files._main_cc[0],
+        output = ctx.outputs.out,
+        substitutions = {
+            "{TargetExe}": exe.out.short_path,
+
+            # Strip the leading "../" off the path to dotnet.exe
+            "{DotnetExe}": toolchain.runtime.short_path[3:],
+
+            "{RuntimeConfigJson}": ctx.file.runtimeconfig.short_path,
+        },
+    )
+
+_write_wrapper_cc = rule(
+    _write_wrapper_cc_impl,
+    doc = "Write a .cc file that will wrap a C# exe",
+    attrs = {
+        "exe": attr.label(
+            doc = "The exe to wrap",
+            providers = AnyTargetFramework,
+        ),
+        "target_framework": attr.string(
+            doc = "The target framework to choose at runtime",
+            mandatory = True,
+        ),
+        "out": attr.output(
+            doc = "The .cc file that gets output",
+        ),
+        "runtimeconfig": attr.label(
+            doc = "A *.runtimeconfig.json file",
+            allow_single_file = True,
+            mandatory = True,
+        ),
+        "_main_cc": attr.label(
+            doc = "The .cc file to use as a template",
+            allow_single_file = [".cc"],
+            default = ":main.cc",
+        ),
+    },
+    toolchains = ["@d2l_rules_csharp//csharp/private:toolchain_type"],
+)
+
+def _csharp_exe_impl(ctx):
+    # Collect the C# providers from the exe
+    csharp_providers = []
+    for tf in CSharpAssembly.values():
+        if tf in ctx.attr.exe:
+            csharp_providers += [ctx.attr.exe[tf]]
+
+    # Get the specific provider for the one we want to run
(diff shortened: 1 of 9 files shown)


csharp/private/tools/assembly-runner/lib.bzl
```diff
@@ -0,0 +1,147 @@
...
+    toolchain = ctx.toolchains["@d2l_rules_csharp//csharp/private:toolchain_type"]
+
+    ctx.actions.expand_template(
+        template = ctx.files._main_cc[0],
+        output = ctx.outputs.out,
+        substitutions = {
+            "{TargetExe}": exe.out.short_path,
+
+            # Strip the leading "../" off the path to dotnet.exe
+            "{DotnetExe}": toolchain.runtime.short_path[3:],
```
j3parker: I think using the toolchain like this might be OK... I'm worried about a situation where our toolchain doesn't match the uh... OS we run `bazel run` on (remote builds?) and we end up picking the wrong one.
We use the toolchain again in `_csharp_exe` with the new `everything` filegroup target.

csharp/private/tools/assembly-runner/lib.bzl
```diff
@@ -0,0 +1,147 @@
...
+
+    # Get the specific provider for the one we want to run
+    provider = CSharpAssembly[ctx.attr.target_framework]
+    exe = ctx.attr.exe[provider]
+
+
+    # TODO: use bat for windows
+    # We need to copy the wrapper because otherwise Bazel responds with this:
+    #  'executable' provided by an executable rule '_csharp_exe' should be created by the same rule.
+    # :(
```
j3parker: This buddy does a cross-platform copy: https://github.com/bazelbuild/bazel-skylib/blob/master/rules/copy_file.bzl

We can't use it directly (because it's a rule, and we want to copy the file within this existing rule specifically).

csharp/private/tools/assembly-runner/main.cc
```diff
@@ -0,0 +1,49 @@
...
+
+  auto dotnet_argv = new char*[argc + 4];
+
+  // dotnet wants this to either be dotnet or dotnet.exe but doesn't have a
+  // preference otherwise.
+  dotnet_argv[0] = (char*)"dotnet";
+  dotnet_argv[1] = (char*)"--runtimeconfig";
+  dotnet_argv[2] = (char*)"{RuntimeConfigJson}";
+
+  dotnet_argv[3] = (char*)exe.c_str();
```
j3parker: For .NET Framework things we probably don't want to wrap with `dotnet`? Not sure if that works.
The wrapping would look similar for #51 I think... I think we could share this template with some complexification.

Eventually for #9 and the assembly loading stuff we'll want to invoke a different C# exe and path this `exe.c_str()` to the first one. The first one will overload assembly resolution to use the runfiles manifest. So I think we will want some form of wrapping no matter what platform/framework, it will just change slightly depending on those things.

PR #89, opened 2019-10-09
disable version smearing for framework libraries

omsmith:


csharp/private/rules/imports.bzl
```diff
@@ -119,14 +124,20 @@ def _import_multiframework_library_impl(ctx):
         if attr != None:
             providers[tfm] = attr[CSharpAssembly[tfm]]
 
-    fill_in_missing_frameworks(providers)
+    if ctx.attr.version_smearing:
+        fill_in_missing_frameworks(providers)
```
omsmith: I kinda feel like `import_multiframework_library` shouldn't ever be doing this? I'm not sure. I'm thinking it's strategy should moreso be "take `CSharpAssembly[][]` and flatten it to `CSharpAssembly[]`, holes are as filled as the providers fill them.

omsmith: Iiii dunno though. It's just our strategy of doing the smearing to get dependency-across-version support coming back to bite us I guess :P

omsmith:
So this on it's own isn't a viable approach. When building a netstandard2.0 library, we try to fill in providers it supports. netstandard2.0 supports netcoreapp2.0, so we try to fill that in. When re-interpreting the dependencies in the context of netcoreapp2.0, we find we don't have any standard references for netcoreapp2.0. That's true, but we also aren't trying to build netcoreapp2.0 so I don't care.

We would run into the same thing for net20 -> net35 I'm sure.

```
/home/omsmith/dev/rules_csharp/examples ~ bazel build ...
ERROR: /home/omsmith/dev/rules_csharp/examples/every_framework_empty/BUILD:81:1: in csharp_library rule //every_framework_empty:netstandard2.0: 
Traceback (most recent call last):
	File "/home/omsmith/dev/rules_csharp/examples/every_framework_empty/BUILD", line 81
		csharp_library(name = 'netstandard2.0')
	File "/home/omsmith/.cache/bazel/_bazel_omsmith/2fb55c7c47f7f9f7347244ac3cb377ea/external/d2l_rules_csharp/csharp/private/rules/library.bzl", line 34, in _library_impl
		fill_in_missing_frameworks(providers)
	File "/home/omsmith/.cache/bazel/_bazel_omsmith/2fb55c7c47f7f9f7347244ac3cb377ea/external/d2l_rules_csharp/csharp/private/common.bzl", line 86, in fill_in_missing_frameworks
		collect_transitive_info(base.deps, tfm)
	File "/home/omsmith/.cache/bazel/_bazel_omsmith/2fb55c7c47f7f9f7347244ac3cb377ea/external/d2l_rules_csharp/csharp/private/common.bzl", line 30, in collect_transitive_info
		fail(("The dependency %s is not compa...)))
The dependency @net//:StandardReferences is not compatible with netcoreapp2.0!
ERROR: Analysis of target '//every_framework_empty:netstandard2.0' failed; build aborted: Analysis of target '//every_framework_empty:netstandard2.0' failed; build aborted
INFO: Elapsed time: 0.068s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

PR #90, opened 2019-10-09
Moving csharp_library to have a single file in returned DefaultInfo

jimevans:
This allows dependent rules to easily use `csharp_library` rules for
things like copying files to another output location. Having multiple
files returned makes such a thing difficult.

### DIFF
--- a/csharp/private/rules/library.bzl
+++ b/csharp/private/rules/library.bzl
@@ -34,7 +34,8 @@ def _library_impl(ctx):
 
     result = providers.values()
     result.append(DefaultInfo(
-        files = depset([result[0].out, result[0].refout, result[0].pdb]),
+        files = depset([result[0].out]),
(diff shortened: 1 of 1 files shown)


j3parker:
@omsmith: Jim and I talked about this in Slack. It's really convenient if `DefaultInfo`s `files` is simple (in this case just the DLL). It makes using our targets from things like `genrule` simpler. Some notes:

* We want to ut things like the pdb and transitive DLLs into runfiles (for `bazel run`/`bazel test`)
* I'll make `csharp_binary` and `csharp_nunit_test` do the same thing.
* More advanced rules (e.g. nuget package creation) should ignore `DefaultInfo` and use the C# providers instead
* `refout` isn't needed in runfiles ever
* We probably want to populate runfiles with transitive deps for libraries too, even though they can't be directly run (other rules might look at their runfiles, though.) I'll deal with that.

PR #91, opened 2019-10-09
Add build badge for Bazel CI

j3parker:


j3parker:
Ironically GitHub actions appears to be having some global issues right now (I've noticed this on some other repos) but Bazel-CI worked! I'm going to switch our branch protection rules over to those checks instead.

Issue #10, opened 2019-09-06
Cross-platform resgen (.resx compilation)


j3parker:
We have been using the system resgen at D2L (not included in this repo). That's Windows-specific and also kind of bad.

It looks like .NET Core doesn't have an equivalent to `resgen.exe` at the moment. It appears to live inside an MSBuild task.

We could add msbuild to our toolchain and generate a stub project to compile resx. Ugly, but I think it'd work. We wouldn't download msbuild unless you use resx so I think this would be fine for now.

jrbeverly:
I'm going to take a look if we are able to use mono's implement of `resgen.exe` to if it would resolve this, or any other options that exist in that space.

Otherwise, I'll take a look about hwat generating a stub project for resx compilation would look like.

PR #93, opened 2019-10-09
WIP: Compile resx files into CLR binary files (.resource)

jrbeverly:
This is a rough outline of the API I am looking at for the rule `csharp_resx`, that is responsible for implementing the behaviors of [resgen](https://docs.microsoft.com/en-us/dotnet/framework/tools/resgen-exe-resource-file-generator). 

This is a stripped down version of my working branch, as I wanted to get some feedback on the direction I am looking at for this.

Relates to #10


csharp/private/rules/Template.csproj
```diff
@@ -0,0 +1,10 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>{FRAMEWORK}</TargetFrameworks>
+        <EnableDefaultEmbeddedResourceItems>False</EnableDefaultEmbeddedResourceItems>
```
jrbeverly: This is to prevent the default compilation includes of .NET projects. If this isn't specified, then any `.resx` file in the same directory as the generated `.csproj` will be included in compilation. I noticed this when testing with localized resource files (e.g. `Strings.resx`, `Strings.en-FR.resx`)

You can see more about this in .NET core projects [here](https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#default-compilation-includes-in-net-core-projects) (I was using `dotnet` for testing this compilation)

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,50 @@
...
+    attrs = {
+        "srcs": attr.label_list(
+            mandatory = True, 
+            allow_files = True
+        ),
+        "csproj": attr.label(
+            mandatory = True, 
+            allow_single_file = True
+        ),
+        "target_frameworks": attr.string_list(
```
jrbeverly: I don't like having to specify the target framework for this, but I noticed when working with localized resources (e.g. `Strings.resx`, `Strings.en-FR.resx`) that dlls were compiled for each localization. I have not had time yet to thoroughly investigate the impact of different framework targets on the compiled CLR binary `.resources`, or the output `.dll` per culture.

For example, for `hello.csproj`, there would be a compiled resources dll in `bin/en-FR/hello.resources.dll`.

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,50 @@
+# Label of the template file to use.
+_TEMPLATE = "@d2l_rules_csharp//csharp/private:rules/Template.csproj"
+_TEMPLATE_EMBEDDED_RESOURCE = "        <EmbeddedResource Include=\"%s\" />"
+
+def _csproj_embedded_resource(resx_files):
+    result = ""
+    for src in resx_files:
+        result += _TEMPLATE_EMBEDDED_RESOURCE % (src.basename)
```
jrbeverly: I'm looking to see if I can figure out a way that uses direct references to the source files as per discussion with @j3parker , as the way I am currently doing this requires copying the `.resx` files into the same directory as the `.csproj`. The issue I am having is that the csproj is looking for the files relative to itself, rather than relative to the workspace.

Experiments with getting the absolute path (or a relative path) from the executing directory of the rule haven't proved very successful.

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,50 @@
...
+    return [DefaultInfo(files = files, runfiles = runfiles)]
+
+csharp_resx = rule(
+    implementation = _csharp_resx_impl,
+    attrs = {
+        "srcs": attr.label_list(
+            mandatory = True, 
+            allow_files = True
+        ),
+        "csproj": attr.label(
```
jrbeverly: I am adding this attribute as a placeholder (it allows it to work in my WIP branch) for what I believe might need to be namespacing information. When using the name of the rule for the `.csproj` (hello.strings.csproj), it resulted in the compiled CLR binary not being discovered when compiling the `csharp_binary`. 

I suspect the reason for that is that the namespace is not set in the `csproj` template. It then must use the name of the  csproj as its namespace. I will be looking into this later when I get a chance.

jrbeverly: I think I can avoid this issue if I make use of [identifier](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/resource-compiler-option) when passing the resource file to msbuild

jrbeverly:
After looking into the satellite assemblies more, I think it'll be easier to work with if the single vs multiple resx are split into two different rules. That should make that API easier to work with when trying to assign identifiers to every resource.

j3parker:
Cool! I'm sorry I didn't get a chance to look at this in detail. I'll wait for the follow-up :)

j3parker:
P.S. a PR with just the API and an empty (`print("todo")`) impl could be a good first step. It will make the individual reviews smaller.

PR #95, opened 2019-10-12
Add rbe_ubuntu1604 to CI

j3parker:
(I think this will fail initially because of implicit unvendored dependencies like LibICU).

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -28,3 +28,12 @@ tasks:
     working_directory: examples
     build_targets:
     - ...
+  examples-rbe_ubuntu1604:
+    name: Examples on the remote build environment
+    platform: rbe_ubuntu1604
+    include_json_profile:
+    - build
+    - test
+    working_directory: examples
+    build_targets:
+    - ...


j3parker:
Hm, the failures aren't the ones I was predicting:

> (16:40:14) ERROR: While resolving toolchains for target //every_framework_empty:netcoreapp3.0: com.google.devtools.build.lib.packages.BuildFileNotFoundException: no such package '@buildkite_config//config': The repository '@buildkite_config' could not be resolved

j3parker:
Cool!

```
A fatal error occurred. The folder [/b/f/w/external/netcore-runtime-linux/host/fxr] does not exist
```

So our action is passing the path to `csc` (the compiler, as an element of `input`) and `dotnet` (the runtime, as the `executable` attr) but not the other files associated with the runtime. I guess the Linux/Mac sandboxes aren't preventing us from reading those extra files, but remote builds caught it. Awesome!

Unfortunately we need to fix that before turning on remote builds!

I still think we're going to hit other issues related to assumptions on libraries that have to exist (see [Microsoft's documentation](https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore30)) but that's less pressing than this bug!

Issue #77, opened 2019-09-27
Need a data structure for resources that allow different display name than file name


jimevans:
Currently the `resources` attribute of `csharp_library` and `csharp_binary` takes a `label_list` of files to be embedded as resources in the resulting assembly. This means that the command line for the compiler turns each label into `/resource:<file name>`. However, that's not the extent of that command line switch, as there are additional options that can be supported. The full range of things that the compiler switch can support are as follows:

    /resource:filename[,identifier[,accessibility-modifier]] 

where `identifier` is an optional name used internally by the application for referring to the resource, and `accessibility-modifier` is either `public` or `private` (defaulting to `public`).

These rules need a mechanism whereby the identifier and accessibility modifier can be optionally specified for each resource. It's unclear to me whether the correct mechanism is a new data structure that one needs to create targets for (like the rules_dotnet's `net_resource` structure), or something else.

jrbeverly:
I should have added to this earlier (when I raised #96). but to make #98 work in a sane way I will be looking to added the provider `CSharpResource`, with the format:

```python
CSharpResource(
    # The name of the bazel rule that generated this resource
    name,
    # The common language runtime binary resource file (.resources).
    result,
    # The logical name for the resource; the name that is used to load the resource. The default is the name of the rule.
    identifier,
)
```

This replaces the `listOf(File)` passed to the compilation action (AssemblyAction). This is still missing the `accessibility-modifier`, instead relying on just defaulting to `public`

PR #96, opened 2019-10-15
WIP: Single resx compilation using dotnet cli for compilation

jrbeverly:
An API for compiling a single resx into a `.resources` binary.

Related to #10

### DIFF
--- a/csharp/private/rules/resgen.bzl
+++ b/csharp/private/rules/resgen.bzl
@@ -1,0 +1,31 @@
+
+def _csharp_resx_impl(ctx):
+    print("todo")
+
+csharp_resx = rule(
+    implementation = _csharp_resx_impl,
+    attrs = {
+        "src": attr.label(
+            doc = "The resx file."
+            mandatory = True, 
+            allow_single_file = True
+        ),
+        "identifier": attr.string(
+            doc = "The identifier of the resource."
+            mandatory = True, 
+        ),
+        "target_frameworks": attr.string_list(
+            doc = "A list of target framework monikers to build.",
+            allow_empty = False,
+        ),
+        "_csproj_template": attr.label(
+            doc = "The csproj template used in compiling a resx file."
+            default = Label(_TEMPLATE),
+            allow_single_file = True,
+        ),
+        "_dotnet_runner": attr.string(
+            doc = "Allows customizing the dotnet instance. [PH]"
+            default = "dotnet",
+        ),
+    },
+)
--- a/examples/resgen/BUILD
+++ b/examples/resgen/BUILD
@@ -1,0 +1,21 @@
+load(
+    "@d2l_rules_csharp//csharp:defs.bzl",
+    "csharp_binary",
+    "csharp_resx"
+)
+
+# bazel run //examples:hello
+csharp_binary(
+    name = "hello",
+    srcs = ["hello.cs", "Strings.Designer.cs"],
+    resources = [":hello.resources"],
+    target_frameworks = ["net472"],
+    deps = ["@net//:System.Linq"],
+)
+
+csharp_resx(
+    name = "hello.resources",
+    src = "Strings.resx",
+    identifier = "Hello.Strings.resource",
+    target_frameworks = ["net472"],
+)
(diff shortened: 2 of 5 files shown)


csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,31 @@
+
+def _csharp_resx_impl(ctx):
+    print("todo")
+
+csharp_resx = rule(
+    implementation = _csharp_resx_impl,
```
jrbeverly: I'm thinking we'll want to provide a custom struct as mentioned in #77, as the name of the output `.resources` file is determined by either the csproj file name or the `RootNamespace` property. We could add an `out` property for modifying the name of the csproj file (rather than relying solely on `identifier`)

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,31 @@
+
+def _csharp_resx_impl(ctx):
+    print("todo")
+
+csharp_resx = rule(
+    implementation = _csharp_resx_impl,
+    attrs = {
+        "src": attr.label(
```
j3parker: It might be more ergonomic for this to be `srcs` and to invoke `dotnet` multiple times in a loop behind the scenes.

examples/resgen/BUILD
```diff
@@ -0,0 +1,21 @@
...
+# bazel run //examples:hello
+csharp_binary(
+    name = "hello",
+    srcs = ["hello.cs", "Strings.Designer.cs"],
+    resources = [":hello.resources"],
+    target_frameworks = ["net472"],
+    deps = ["@net//:System.Linq"],
+)
+
+csharp_resx(
```
j3parker: Looks good. If `src` -> `srcs` then it'd be idiomatic to name this rule `csharp_resx_library`.

csharp/private/rules/resgen.bzl
jrbeverly: The struct containing a `identifier` property (or `out` for controlling output name) is not really related to the csproj, but to how the resources are consumed by [C# compiler](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/resource-compiler-option):

```
-resource:filename[,identifier[,accessibility-modifier]]
```

With parameters like that, I can pass in a resources file named `hello.look.at.this.name.resources`, but use the identifier `hello.Strings.resources`. This is what I was getting at in [this](https://github.com/Brightspace/rules_csharp/pull/93#discussion_r333149380) discussion

csharp/private/rules/resgen.bzl
jrbeverly: If we went this route, the API for specifying the `identifier` would need to be a [label_keyed_string_dict](https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_keyed_string_dict), which makes the specification of the `identifier` mandatory.

I would also have concerns with the resource being used incorrectly with localized resx files. For localized resx files, satellite assemblies are created in directories based on the culture code. Those will need some special handling, as the pathing of those assemblies matter.

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,31 @@
...
+
+csharp_resx = rule(
+    implementation = _csharp_resx_impl,
+    attrs = {
+        "src": attr.label(
+            doc = "The resx file."
+            mandatory = True, 
+            allow_single_file = True
+        ),
+        "identifier": attr.string(
```
jrbeverly: I've been looking into an `out` parameter to specify the name of the compiled resource.  With `EmbeddedResource`'s you can specify the property `LogicalName` that is used by [GenerateResource](https://docs.microsoft.com/en-us/visualstudio/msbuild/generateresource-task?view=vs-2019) to set the output name.

> The GenerateResource task uses the <LogicalName> metadata of an <EmbeddedResource> item to name the resource that is embedded in an assembly.
>
> Assuming that the assembly is named myAssembly, the following code generates an embedded resource named someQualifier.someResource.resources:

Based on my testing so far I haven't been successful with getting the compiled resource named based on the `LogicalName` property. I'll try testing this from visual studio, to get a better idea of how this is happening in that environment.

jrbeverly:
Closing draft as working copy is now raised at #98.

PR #97, opened 2019-10-23
Replace dotnet runtime with SDK

jrbeverly:
ResX compilation works by creating a `.csproj` with the resx file as an embedded resource. When compiled (using GenerateResourceTask), it will create the `.resources` file that can then be embedded in the csharp lib/binary. To compile the csproj, we need the dotnet sdk.

Downloading the dotnet SDK on windows (`dotnet-sdk-windows`) takes around 86 seconds, and goes from ~45 MB (runtime + compiler) to ~110-150 MB (depending on OS).

Related to #10 and #96


j3parker:
@omsmith you cool with this?

PR #98, opened 2019-10-23
Compile resx file into binary resource file for use in AssemblyAction

### LINKED ISSUE #10: Cross-platform resgen (.resx compilation)

jrbeverly:
Compiles a single resx file into a common language runtime binary resource file.

Example usage:
```bazel
csharp_resx(
    name = "hello.resources.mystrings",
    src = "Strings.resx",
    identifier = "Hello.Strings.resources",
)
```

There a couple of things I'll mention here:

- ResX files have a lot of cases that I think should be tested (multiple resx files, embedded files, identifier/out parameters, etc)
- A target_framework is necessary for the csproj to compile

Resolves #10 , #77


csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,99 @@
...
+    args.add(csproj.path)
+    
+    ctx.actions.run(
+        inputs = [ctx.file.src, csproj],
+        outputs = [resource],
+        executable = toolchain.runtime,
+        arguments = [args],
+        mnemonic = "CompileResX",
+        progress_message = "Compiling resx file to binary",
+        env = {
```
jrbeverly: Without these 4 environment variables, `dotnet` cli just won't run. @j3parker, do you have any ideas on how I might be able to: 1) not have to specify these 2) specify these for each OS

While they are set to `/root/`, they will fail on the linux & macOS builds (as they wouldn't have permission to create `/root/.dotnet`). Additionally without `DOTNET_CLI_TELEMETRY_OPTOUT=1`, the telemetry message will output a bunch.

examples/resgen/Strings.Designer.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Hello {
```
omsmith: Should this have been committed?

examples/resgen/BUILD
```diff
@@ -0,0 +1,18 @@
...
+    srcs = ["hello.cs", "Strings.Designer.cs"],
+    resources = [":Hello.Strings"],
+    target_frameworks = ["net472"],
+    deps = ["@net//:System.Linq"],
+)
+
+csharp_resx(
+    name = "Hello.Strings",
+    src = "Strings.resx",
+)
```
omsmith: nit: Missing trailing newlines on a number of files

csharp/private/rules/resgen.bzl
```diff
@@ -0,0 +1,99 @@
...
+
+# Label of the csproj template for ResX compilation
+_TEMPLATE = "@d2l_rules_csharp//csharp/private:rules/ResGen.csproj"
+
+# When compiling the csproj, it will look for the embedded resources related to
+# the path of the csproj on disk. Since the csproj is written by the bazel, it will
+# be in the rule bin directory. If we want to get the relative path to the
+# ResX files, we need to get to a junction/symlink for the source. 
+#
+# The junction/symlink that we are using is 4 directories up (\bazel-out\<sandbox>\bin\<name>\)
```
omsmith: Relying on this seems, likely to break.

jrbeverly: Agree, but I have had difficulty finding a way of configuring relative paths for the `.resx` files that isn't a bit 'hacky'.

Ideally I'd just go with the copy route, but that doesn't handle embedded files like so:

```xml
  <data name="SetupFiles" type="System.Resources.ResXFileRef, System.Windows.Forms">
    <value>Resources\SetupFiles.zip;System.Byte[], mscorlib, Version=4.0.0.0, ...</value>
  </data>
```

examples/resgen/Strings.Designer.cs
jrbeverly: The `.Designer.cs` file? Yes. Without it you would just get a compilation error about `Strings` not existing in the context.

You can generate this file in the compilation process using a `Compile` task on the designer file (e.g. `<Compile Update="Strings.Designer.cs">`).

csharp/private/rules/resgen.bzl
jrbeverly: Getting back into this after inspiration, I think the best option for now would be a wrapper over `dotnet` that ensures the appropriate variables are passed in for each execution. This way, we don't have to worry about the `.dotnet` directory of one rule conflicting with that of another.

I looked into other options, but they were mostly dead ends.

jrbeverly:
I have split out the work for running `dotnet` in another pull request #102. Using some of the work I did there with runfiles, I will be revising how the resx files are compiled in this pull request. At the moment I use a pathing hack (`../../../../`) to try and find the `resx` file. This only works for files that are one directory below the WORKSPACE (overfitting of the examples). 

The new approach I'll be using the runfiles data to create the csproj file. This will use an absolute path for the resx file (ensuring all embedded files are working as intended), and convert the rule to a macro.

I'll look to raise a pull request tomorrow that uses this new approach for compilation of resx files.

jrbeverly:
I have managed to get a solution working with the runfiles (and learned a fair bit about runfiles). You can see a preview of what is looks like [here](https://github.com/Brightspace/rules_csharp/compare/master...jrbeverly:proto/resx-macro-tangent-2?expand=1). I'll be including the learnings from that branch into this one, as that solution is a bit wonky.

I'll be trimming it down and including the changes into this pull request.

jrbeverly:
I have simplified the work done down to two rules, and moved the C++ compilation wrapper into the bash wrapper. This wrapper doesn't have as much going on (besides the boilerplate runfiles code).

I did discover something while working on windows, and that is that there exists a segmentation fault in the `dotnetw` wrapper. So I'll creating an issue to try to resolve that. I suspect it is related to the `dotnet_argv` that are passed in from `argv`. I suspect it needs to be copied rather than just reference.

j3parker:
Sorry in the delay getting back to this, I'll go through it on the weekend when I have some quiet time :)

jrbeverly:
While doing the housekeeping on some of these PRs, I noticed that with the latest release of the dotnet SDK `3.1.100`, it introduced the expectation for `project.assets.json`.

That was failing the resx compilation due to its usage of `dotnet build`. Based on a quick search, I suspect this'll impact the current implementation of resx.

omsmith:
:(

PR #102, opened 2019-11-21
Wrap the dotnet executable to set default environment variables

jrbeverly:
While working on #98 I encountered difficulty while trying to run the `dotnet.exe` cross-platform. It has the requirement that the environment variables `HOME, DOTNET_CLI_HOME, APPDATA, PROGRAMFILES` are defined. 

Although constants could works (conditional by OS), this means that properties of a bazel dir could impact how a build compiles. Instead, this PR creates a wrapper `dotnetw` that sets environment variables before execution.

This is a draft at the moment while I iron about some edge cases (and refactor the wrapper)

Related to #10

### DIFF
--- a/csharp/defs.bzl
+++ b/csharp/defs.bzl
@@ -8,6 +8,10 @@ load(
     _csharp_binary = "csharp_binary",
 )
 load(
+    "//csharp/private:rules/wrapper.bzl",
+    _dotnet_wrapper = "dotnet_wrapper",
+)
+load(
     "//csharp/private:rules/library.bzl",
     _csharp_library = "csharp_library",
 )
@@ -34,6 +38,7 @@ load(
     _setup_basic_nuget_package = "setup_basic_nuget_package",
 )
 
+dotnet_wrapper = _dotnet_wrapper
 csharp_binary = _csharp_binary
 csharp_library = _csharp_library
 csharp_library_set = _csharp_library_set
--- a/csharp/private/BUILD
+++ b/csharp/private/BUILD
@@ -1,13 +1,10 @@
 load(":toolchains.bzl", "configure_toolchain")
 
-exports_files(["nunit/shim.cs"])
+exports_files(["nunit/shim.cs", "rules/ResGen.csproj", "rules/main.cc"])
 
 toolchain_type(name = "toolchain_type")
 
-configure_toolchain(
-    "windows",
-    exe = "dotnet.exe",
-)
+configure_toolchain("windows")
 
 configure_toolchain("linux")
 
--- a/csharp/private/BUILD
+++ b/csharp/private/BUILD
@@ -1,6 +1,6 @@
 load(":toolchains.bzl", "configure_toolchain")
 
-exports_files(["nunit/shim.cs", "rules/ResGen.csproj", "rules/main.cc"])
+exports_files(["nunit/shim.cs", "wrappers/dotnet.cc"])
 
 toolchain_type(name = "toolchain_type")
 
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -153,6 +153,8 @@ def AssemblyAction(
     # spill to a "response file" when the argument list gets too big (Bazel
     # makes that call based on limitations of the OS).
     args.set_param_file_format("multiline")
+    # Our wrapper uses _spawnv to launch dotnet, and that has a command line limit
+    # of 1024 bytes, so always use a param file.
     args.use_param_file("@%s", use_always=True)
 
     direct_inputs = srcs + resources + analyzer_assemblies + additionalfiles + [toolchain.compiler]
--- a/csharp/private/actions/assembly.bzl
+++ b/csharp/private/actions/assembly.bzl
@@ -153,7 +153,7 @@ def AssemblyAction(
     # spill to a "response file" when the argument list gets too big (Bazel
     # makes that call based on limitations of the OS).
     args.set_param_file_format("multiline")
-    args.use_param_file("@%s")
+    args.use_param_file("@%s", use_always=True)
 
     direct_inputs = srcs + resources + analyzer_assemblies + additionalfiles + [toolchain.compiler]
     direct_inputs += [keyfile] if keyfile else []
--- a/csharp/private/rules/wrapper.bzl
+++ b/csharp/private/rules/wrapper.bzl
@@ -1,0 +1,30 @@
+def _dotnet_wrapper_impl(ctx):
+    cc_file = ctx.actions.declare_file("%s.cc" % (ctx.attr.name))
+    ctx.actions.expand_template(
+        template = ctx.file.src,
+        output = cc_file,
+        substitutions = {
+            "{DotnetExe}": ctx.files.target[0].short_path[3:],
+        },
+    )
+
+    files = depset(direct = [cc_file])
+    return [
+        DefaultInfo(
+            files = files,
+        ),
+    ]
+
+dotnet_wrapper = rule(
+    implementation = _dotnet_wrapper_impl,
+    attrs = {
+        "src": attr.label(
+            mandatory = True, 
+            allow_single_file = True
+        ),
+        "target": attr.label_list(
+            mandatory = True, 
+            allow_files = True,
+        ),
+    },
+)
--- a/csharp/private/rules/wrapper.bzl
+++ b/csharp/private/rules/wrapper.bzl
@@ -1,10 +1,15 @@
+_TEMPLATE = "@d2l_rules_csharp//csharp/private:wrappers/dotnet.cc"
+
 def _dotnet_wrapper_impl(ctx):
     cc_file = ctx.actions.declare_file("%s.cc" % (ctx.attr.name))
+    if len(ctx.files.src) < 1:
+        fail("No dotnet executable found in the SDK")
+
     ctx.actions.expand_template(
-        template = ctx.file.src,
+        template = ctx.file.template,
         output = cc_file,
         substitutions = {
-            "{DotnetExe}": ctx.files.target[0].short_path[3:],
+            "{DotnetExe}": ctx.files.src[0].short_path[3:],
         },
     )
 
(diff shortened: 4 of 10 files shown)


csharp/private/actions/assembly.bzl
```diff
@@ -153,7 +153,7 @@ def AssemblyAction(
     # spill to a "response file" when the argument list gets too big (Bazel
     # makes that call based on limitations of the OS).
     args.set_param_file_format("multiline")
-    args.use_param_file("@%s")
+    args.use_param_file("@%s", use_always=True)
```
j3parker:
```suggestion
    # Our wrapper uses _spawnv to launch dotnet, and that has a command line limit
    # of 1024 bytes, so always use a param file.
    args.use_param_file("@%s", use_always=True)
```

csharp/private/BUILD
```diff
@@ -1,13 +1,10 @@
 load(":toolchains.bzl", "configure_toolchain")
 
-exports_files(["nunit/shim.cs"])
+exports_files(["nunit/shim.cs", "wrappers/dotnet.cc"])
```
j3parker: Why is this being exported?

csharp/private/runtime.BUILD
```diff
@@ -1,9 +1,14 @@
...
-        "dotnet.exe",  # windows, yeesh
-    ], allow_empty = True) +
-    glob([
+    glob(
+        [
+            "dotnet",
+            "dotnet.exe",  # windows, yeesh
+        ],
+        allow_empty = True,
+    ) + glob([
```
j3parker: What's this change about?

csharp/private/runtime.BUILD
```diff
@@ -13,3 +18,26 @@ exports_files(
...
 )
+
+cc_binary(
+    name = "dotnetw",
+    srcs = [":main-cc"],
+    data = glob(
+        [
+            "dotnet",
+            "dotnet.exe",
+        ],
```
j3parker: Could you run this through buildifier?

j3parker: I think this will need `allow_empty = True` (to pass with `--strict`)

csharp/private/wrappers/dotnet.cc
```diff
@@ -0,0 +1,104 @@
...
+#else // not _WIN32
+#include <stdlib.h>
+#include <unistd.h>
+#endif // _WIN32
+
+#include "tools/cpp/runfiles/runfiles.h"
+
+using bazel::tools::cpp::runfiles::Runfiles;
+
+std::string evprintf(std::string name, std::string path)
```
j3parker: Can you run this through clang-format with the Google style rules?

csharp/private/BUILD
jrbeverly: The file `dotnet.cc` is used for wrapping the `dotnet` executable, you can see the template reference [here](https://github.com/Brightspace/rules_csharp/blob/cb981c78d77bd79887c19e4fdab9bbece967778f/csharp/private/rules/wrapper.bzl#L1).

csharp/private/runtime.BUILD
jrbeverly: This is due to buildifier changing the formatting of this file. I verified that this changed on both my branch, and running on master for this file. In both cases, the output result is as above.

csharp/private/runtime.BUILD
jrbeverly: It does have an issue with empty globs when running `bazel build --incompatible_disallow_empty_glob :*`. I'll add the `allow_empty` flag to these usages.

If anyone is interested in the list, it is [here](https://docs.bazel.build/versions/master/command-line-reference.html#flag--incompatible_disallow_empty_glob)

csharp/private/wrappers/dotnet.cc
jrbeverly: I have added `.clang-format` to the wrappers/ directory so that any new wrappers will follow the format guidelines.

csharp/defs.bzl
```diff
@@ -34,6 +38,7 @@ load(
     _setup_basic_nuget_package = "setup_basic_nuget_package",
 )
 
+dotnet_wrapper = _dotnet_wrapper
```
j3parker: Why is this being exported? Shouldn't this be internal to these rules?

csharp/private/wrappers/.clang-format
```diff
@@ -0,0 +1,14 @@
+# Use the Google style in this project.
+BasedOnStyle: Google
+
+DerivePointerAlignment: false
```
j3parker: What's this?

csharp/private/wrappers/dotnet.cc
```diff
@@ -0,0 +1,97 @@
...
+
+  // Get the name of the directory containing dotnet.exe
+  auto dotnetDir = dotnet.substr(0, dotnet.find_last_of("/\\"));
+
+  /*
+  dotnet and nuget require these environment variables to be set
+  without them we cannot build/run anything with dotnet.
+
+  dotnet: HOME, DOTNET_CLI_HOME, APPDATA, PROGRAMFILES
+  nuget: TMP, TEMP, USERPROFILE
```
j3parker: We don't need to run `nuget` though?

csharp/private/wrappers/dotnet.cc
```diff
@@ -0,0 +1,97 @@
...
+  */
+  std::vector<std::string> envvars;
+  envvars.push_back(evprintf("HOME", dotnetDir));
```
j3parker: Use list initialization:

```
std::vector<std::string> envvars = {
  "foo",
  "bar",
  ...
}
```

csharp/private/wrappers/dotnet.cc
```diff
@@ -0,0 +1,97 @@
...
+  dotnet_argv[0] = (char*)"dotnet";
+  for (int i = 1; i < argc; i++) {
+    dotnet_argv[i] = argv[i];
+  }
+  dotnet_argv[argc] = nullptr;
+
+#ifdef _WIN32
+  // _spawnve has a limit on the size of the environment variables
+  // passed to the process. So here we will set the environment
+  // variables for this process, and the spawned instance will inherit them
```
j3parker: wtf lol

csharp/defs.bzl
j3parker: I see, you are using it inside the runtime BUILD

csharp/private/wrappers/.clang-format
jrbeverly: >DerivePointerAlignment (bool)
> If true, analyze the formatted file for the most common alignment of & and *. Pointer and reference alignment styles are going to be updated according to the preferences found in the file. PointerAlignment is then used only as fallback.

Its for cases where people are writing `int& foo` vs `int &foo`. Since we are following google guidelines, they only ask for consistency in a project. Setting this to false should enforce `int& foo` in the project.

csharp/private/wrappers/dotnet.cc
jrbeverly: For commit 80e56d3d468ac767c98b22231df22891057d25aa I tried switching over to list initialization, but MacOS didn't love that. I suspect I might need to cast every single list item into a `const_char` to get it to work.

csharp/private/wrappers/dotnet.cc
jrbeverly: It isn't really about usage (as nothing in the resgen uses nuget), and more a matter of the `dotnet` executable establishing base assumptions. If we don't ensure those variables are accessible, then:

- `dotnet.exe` will reach out to `C:\WINDOWS\NuGetScratch\lock` to see if a lock is held on nuget
- `dotnet.exe` doesn't have permissions to access `C:\WINDOWS\Temp`, so access error
- Any error (including access) is seen as the lock being held already
- Since the lock is held, `dotnet.exe` errors, recommending you wait / release the lock

It may also look in the `AppData/Local` directory for the lock.

csharp/private/wrappers/dotnet.cc
jrbeverly: 🥇

csharp/private/wrappers/dotnet.cc
jrbeverly: Setting `--no-restore` and `--no-dependencies` may prevent the need for those environment variables.

csharp/private/wrappers/dotnet.cc
jrbeverly: Looks like that was it

csharp/private/wrappers/dotnet.cc
jrbeverly: The nuget environment variables here were needed for the resx code, so I'm going to remove them from here so they can be included in the resx PR if necessary.

jrbeverly: Seems that `USERPROFILE` is needed for MacOS, so I will be adding that back. It looks like only the `TMP` variables were special for the resx code.

apriyadarsheeD2L:
Were you able to do your testing on Mac?

jrbeverly:
Yes, for commit 8d0ef05ff1a94a3423fb18be05f1b1293367a8d2 I changed to list initialization and enforced `const_char` on the string values. That resolved the issue I was seeing on MacOS.

Seems that the MacOS version of `execve` has strict expectations with the inputs being const.

csharp/private/rules/wrapper.bzl
```diff
@@ -0,0 +1,39 @@
...
+dotnet_wrapper = rule(
+    implementation = _dotnet_wrapper_impl,
+    attrs = {
+        "template": attr.label(
+            doc = """Path to the program that will wrap the dotnet executable.
+This program will be compiled and used instead of directly calling the dotnet executable.""",
+            default = Label(_TEMPLATE),
+            allow_single_file = True
+        ),
+        "src": attr.label_list(
```
j3parker: nit: this file isn't really used at compile time so it's a tiny bit weird to call it `src`, but that's fine. This is an internal rule.

csharp/private/wrappers/dotnet.cc
```diff
@@ -0,0 +1,93 @@
...
+    std::cerr << "Couldn't find the .NET runtime" << std::endl;
+    return 404;
+  }
+
+  // Get the name of the directory containing dotnet.exe
+  auto dotnetDir = dotnet.substr(0, dotnet.find_last_of("/\\"));
+
+  /*
+  dotnet requires these environment variables to be set.
+  */
```
j3parker: nit: inconsistent comment style in this file

PR #108, opened 2019-12-05
Resolve segmentation error with dotnet wrapper

### LINKED ISSUE #107: Dotnet wrapper encounters segmentation faults during execution
When running bazel on Windows, there is a chance that the dotnet wrapper `dotnetw` will throw a segmentation fault. The error looks like something like this:

```bash
line 60:  2279 Segmentation fault      ${dotnet_path} build $@ ${CsProjFile}
```

This only seems to be happening on the branch for ResX, as that makes use of a different build scenario.

**Steps to Reproduce**
You can reproduce the segmentation fault while on the branch for #98 by running the following:

```bash
bazel clean
bazel build //resgen:hello
```

If it does not occur immediately, you should be able to repeat the command to eventually receive it. 

**Notes**
This was encountered during the work for #98.

I suspect the issue is caused by the windows command `_spawnv`, as it may have special requirements similar to MacOS. Rather than using pointers to `argv`, I would guess that we need to copy the data to a new array before passing to `_spawnv`.

I have mainly encountered it while working on the ResX branch, which uses `dotnetw build`, so that may be related to this work (as `dotnet build` relies on environment vars more than just `dotnet`)

jrbeverly:
The issue was an incorrect array size for the copied arguments. The array was set to size `argc`, but then an element outside of those bounds was being setting `nullptr`.

Tested by running the code over and over and over:
```bash
for i in {1..15}; do bazel clean && bazel build ///resgen:hello; done
```

Resolves #107 

I'll sum up my emotions as such:
image

### DIFF
--- a/csharp/private/wrappers/dotnet.cc
+++ b/csharp/private/wrappers/dotnet.cc
@@ -52,7 +52,7 @@ int main(int argc, char** argv) {
 
   // dotnet wants this to either be dotnet or dotnet.exe but doesn't have a
   // preference otherwise.
-  auto dotnet_argv = new char*[argc];
+  auto dotnet_argv = new char*[argc + 1];
   dotnet_argv[0] = (char*)"dotnet";
   for (int i = 1; i < argc; i++) {
     dotnet_argv[i] = argv[i];


j3parker:
lol opps

Issue #72, opened 2019-09-26
Add rule to create self-contained binaries for .NET Core


j3parker:
`csharp_binary` outputs IL files ("DLLs", but we currently give them a `.exe` extension anyway) but people will probably want the ability to create self-contained executable files.

With the usual SDK, you use `dotnet publish --self-contained` (see [the docs](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish?tabs=netcore21)).

I don't think this should be the default, because it would be simpler to keep everything in DLLs within the build graph (e.g. so you can add the output of `csharp_binary` to the `deps` of another rule.. a common use case being tests). I think we should have a specific rule for "packaging" a binary up. These rules wouldn't emit `CSharpAssembly_*` providers.

We need to investigate how this stuff works more. We don't have access to `dotnet publish` directly because we don't have the SDK. The SDKs are framework-specific, and right now we are manually using the runtime + the compiler as our toolchain (rather than having one toolchain per (platform, framework) pair).

jrbeverly:
As of #97, we now have access to the full SDK. So for this task we now have access to `dotnet publish` directly.

For publishing up a self-contained binary I suspect we would have to generate a csproj file (similar to ResX). This would also likely be the case for #30.

There is discussion of this nature [for dotnet pack](https://github.com/NuGet/Home/issues/4254#issuecomment-478633882) about needing to generate a stub csproj file.

PR #109, opened 2019-12-10
Use pathmap for cross-machine determinism in PDBs

### LINKED ISSUE #13: Use pathmap for cross-machine determinism
This is at least required for PDBs, as [documented here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/pathmap-compiler-option) but I also noticed absolute paths in our DLLs.

We can't trivially do this in Bazel. We need to figure out all of our command line flags during the analysis phase, which doesn't (afaik) know where the build will actually execute (it could be on a remote machine, and analysis also gets cached) so we can't know what paths to map.

We could wrap the compiler to inject this argument at runtime.

We're talking about adding a wrapper already in #12, but I have some concerns. That thing itself needs to be compiled, and probably also needs `pathmap` to be deterministic, so we've got an even worse chicken and egg problem. Maybe when we compile the wrapper we need a small C program to inject this argument?

I think we need to be flexible and experiment to find out what will work best. The experiments should look at how local and remote caching behave.

jrbeverly:
The source path for files in embedded into the '.pdb' file that is created as part of the build process. You can see this in `hello.pdb` when building the [basic](examples/basic) example . If we want to have deterministic builds, these paths cannot be absolute and linking into the bazel sandbox.

Here is what a (cleaned up) embedded path looks like:

```
root/.cache/bazel/_bazel_root/1a06c90be1eefea8b933cc8429eff806/sandbox/processwrapper-sandbox/28/execroot__main__/basic/hello.cs
```

This PR sets the `/pathmap` property on compilation, using variable substitution in the dotnet wrapper to provide context of the path location. This requires parsing all of the arguments and params file to make the necessary variable substitutions.

The source path is replaced with a constant prefix based on the default path separator (`/ \\`). 

Resolves #13


csharp/private/wrappers/dotnet.cc
```diff
@@ -1,26 +1,99 @@
...
   ss << name << "=" << path;
   return ss.str();
 }
 
+/**
+ * Returns the full path of the current working directory.
+ *
+ * @return Full path of the current working directory. Empty if cannot be set.
+ */
+std::string PWD() {
```
j3parker: https://en.cppreference.com/w/cpp/filesystem/current_path

csharp/private/actions/assembly.bzl
```diff
@@ -102,6 +102,17 @@ def AssemblyAction(
...
+    # The full path of source is embedded into PDBs when compiled
+    # To resolve this, we use pathmap to substitute in new paths
+    # that ensure that these are deterministic
+    #
+    # The dotnet wrapper (dotnetw) provides pathing variables
+    # that can be substituted into any arguments passed in
+    # These are of the form: __BAZEL_{name}_
+
+    # pathmap
+    args.add("/pathmap:%s=%s" % ("__BAZEL_SANDBOX__", "__BAZEL_WORKSPACE__"))
```
j3parker: Why pass this arg at all (and why the string formatting?)

csharp/private/wrappers/dotnet.cc
```diff
@@ -1,26 +1,99 @@
...
+
+/**
+ * Replaces keyword items in a parameters file using the substitutions
+ * dictionary.
+ *
+ * @param file The path to the parameters file.
+ * @param subst Substitutions to make when replacing keywords.
+ * @return true if the substitutions were made; false otherwise.
+ */
+bool ResolveParamsFile(std::string file, SubstitutionMap subst) {
```
j3parker: Yeah, so if we don't pass `/pathmap` in from Bazel we could just append it to argv rather than reaching into the file :)

There is the spawnv limit but since Windows paths are limited anyway it seems like it'd be guaranteed to fit along with the rest of our argv.

csharp/private/wrappers/dotnet.cc
```diff
@@ -50,6 +123,29 @@ int main(int argc, char** argv) {
...
+#ifdef _WIN32
+      {"__BAZEL_WORKSPACE__", "\\"},
+#else   // not _WIN32
+      {"__BAZEL_WORKSPACE__", "/"},
+#endif  // _WIN32
+      {"__BAZEL_SANDBOX__", workspaceDir},
+  };
+
+  // variable substitution of inputs + params file
+  for (int i = 1; i < argc; i++) {
```
j3parker: We'll lose a lot of loops by just appending `/pathmap` rather than searching for it. It looks like we'd also support multiple of these kind of arguments but the motivation for that isn't clear to me.

csharp/private/actions/assembly.bzl
jrbeverly: I wrote this line with the idea in mind that we may need multiple `/pathmap` statements, which would be something like: `args.add_all(pathmaps, map_each = _format_pathmap_arg)`. Although I have checked with our examples, I'll likely take a look at a couple more cases to see if we need anymore statements.

The processor variables (`__BAZEL_<>_`) are so the wrapper can substitute context variables (e.g. filesystem paths) into the command line parameters.

The strings are constant, and resolve to the following: `/pathmap:__BAZEL_SANDBOX__=__BAZEL_WORKSPACE__`.

csharp/private/wrappers/dotnet.cc
jrbeverly: I assume you mean using an environment variable as an action switch (`BAZEL_ACTION=ASSEMBLY_BUILD`), which would conditionally append `/pathmap` to the arguments?

Depending on the implementation of #10 & #72 , we may be making use of other commands of `dotnet` (publish / build /. etc), which wouldn't support the `/pathmap` argument. So to append that argument the wrapper would need to understand the command that is being executed.

I thought using variable substitutions would allow the least amount of friction in this case, as it wouldn't require the wrapper to understand the commands we executed, just how to do `${MY_VAR}` in a cross-platform way.

csharp/private/wrappers/dotnet.cc
jrbeverly: I don't expect to support more variables, just those that help scrub away baked in context like paths.

csharp/private/wrappers/dotnet.cc
jrbeverly: As an addon to the above, the ideas that sort of pushed me towards using variable substitution were as follows:

- I don't want the `dotnet` wrapper to try and understand the commands I am executing
- I want to scrub away context `dotnet` requires in the build process (e.g. paths) 
- I want all the configuration about options/parameters/commands to remain in bazel

csharp/private/actions/assembly.bzl
j3parker: Ah, I can't think of a scenario where we'd pass multiple `/pathmap`. It's something that `dotnet.exe` supports but I don't think we'd have a use-case (for us we just need determinism, and one pathmap should be sufficient for that).

csharp/private/wrappers/dotnet.cc
j3parker: > I assume you mean using an environment variable as an action switch (BAZEL_ACTION=ASSEMBLY_BUILD), which would conditionally append /pathmap to the arguments?

I'd prefer it if we just did it always for now, rather than conditionally.

> Depending on the implementation of #10 & #72 , we may be making use of other commands of dotnet (publish / build /. etc), which wouldn't support the /pathmap argument. So to append that argument, the wrapper would need to understand the command that is being executed.

Ok, let's cross that bridge when it comes to it. We can always have a single boolean switch that we pass to the wrapper if we need to, rather than the substitution thing. But let's not do anything prematurely.

> I thought using variable substitutions would allow the least amount of friction in this case, as it wouldn't require the wrapper to understand the commands we executed, just how to do ${MY_VAR} in a cross-platform way.

I think it's fine for the wrapper to understand this. It's all in the same repo and the wrapper is useless outside of this repo. Let's just keep it simple while we can.

csharp/private/wrappers/dotnet.cc
j3parker: So all I'm suggesting is to replace this with `"/pathmap:" + workspaceDir + "=" + dirSeperator`. The only extra thing the exe knows about is how `/pathmap` is formatted which seems perfectly fine: the job of the wrapper is to execute `dotnet.exe` inside Bazel and that's just part of it.

csharp/private/wrappers/dotnet.cc
jrbeverly: We are at that bridge with the current proposed implementation for compilation of resource files (see #98). The internals of that PR are using `dotnet build <csproj>` to compile resource files using the dotnet executable. This is the case I was describing above.

PR #110, opened 2019-12-10
Generate the APIReference using Stardoc

### LINKED ISSUE #17: Generate an API reference page with Stardoc
Use [stardoc](https://github.com/bazelbuild/skydoc) to generate [docs/APIReference.md](docs/APIReference.md) and check this in.

We can compare the generated version to the checked in version in CI.

jrbeverly:
Add support for stardoc into our bazel rules. This allows us to build the docs using `bazel build //docs/...` or using the helper script `update-docs.sh`.

This PR adds the ability to compile an API Reference for our bazel rules, based on the index for docs (`csharp/defs.for_docs.bzl`). Splitting this from the usage index (`defs.bzl`) helps avoid mixing the docs and compilation based actions, which at the moment is useful as `providers.bzl` is causing some issues.

I have noticed an issue where any file that relies (loads) `providers.bzl` cannot be compiled with stardoc. Although it can be compiled on its own (if the header comment is removed), I think it'll require a bit more investigation to determine exactly why this isn't working as intended.

Resolves #17

### DIFF
--- a/csharp/BUILD
+++ b/csharp/BUILD
@@ -1,16 +1,6 @@
-load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-
-package(default_visibility = ["//visibility:public"])
-
-bzl_library(
-    name = "bzl",
-    srcs = [
-        "defs.for_docs.bzl",
-    ],
-    visibility = ["//visibility:public"],
-    deps = [
-        "//csharp/private:bzl",
-    ],
+exports_files(
+    ["defs.for_docs.bzl"],
+    visibility = ["//docs:__pkg__"],
 )
 
 filegroup(
@@ -18,9 +8,5 @@ filegroup(
     srcs = glob(["*.bzl"]) + [
         "//csharp/private:bzl_srcs",
     ],
-    visibility = ["//visibility:public"],
-    # visibility = [
-    #     "//csharp:__pkg__",
-    #     "//docs:__pkg__",
-    # ],
+    visibility = ["//docs:__pkg__"],
 )
--- a/csharp/BUILD
+++ b/csharp/BUILD
@@ -6,7 +6,7 @@ exports_files(
 filegroup(
     name = "bzl_srcs",
     srcs = glob(["*.bzl"]) + [
-        "//csharp/privates:bzl_srcs",
+        "//csharp/private:bzl_srcs",
     ],
     visibility = ["//docs:__pkg__"],
 )
--- a/csharp/BUILD
+++ b/csharp/BUILD
@@ -6,7 +6,7 @@ exports_files(
 filegroup(
     name = "bzl_srcs",
     srcs = glob(["*.bzl"]) + [
-        "//csharp/private:bzl_srcs",
+        "//csharp/privates:bzl_srcs",
     ],
     visibility = ["//docs:__pkg__"],
 )
--- a/csharp/BUILD
+++ b/csharp/BUILD
@@ -12,3 +12,15 @@ bzl_library(
         "//csharp/private:bzl",
     ],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["*.bzl"]) + [
+        "//csharp/private:bzl_srcs",
+    ],
+    visibility = ["//visibility:public"],
+    # visibility = [
+    #     "//csharp:__pkg__",
+    #     "//docs:__pkg__",
+    # ],
+)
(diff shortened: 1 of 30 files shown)


jrbeverly:
This is a [preview](https://github.com/jrbeverly/rules_csharp/blob/jb/stardoc-rules/docs/APIReference.md) of what the rendered markdown looks like for the stardoc.

csharp/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
+package(default_visibility = ["//visibility:public"])
+
+bzl_library(
+    name = "bzl",
+    srcs = [
+        "defs.for_docs.bzl",
+    ],
+    visibility = ["//visibility:public"],
```
j3parker: Does this need to be public?

csharp/private/BUILD
```diff
@@ -1,6 +1,23 @@
 load(":toolchains.bzl", "configure_toolchain")
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
-exports_files(["nunit/shim.cs", "wrappers/dotnet.cc"])
+package(default_visibility = ["//visibility:public"])
```
j3parker: Does this need to be public?

csharp/private/actions/BUILD
```diff
@@ -0,0 +1,9 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
+package(default_visibility = ["//visibility:public"])
```
j3parker: I'm nervous about making our private stuff public (see the path)

docs/APIReference.md
```diff
@@ -1,3 +1,116 @@
-# API Reference
+<!-- Generated with Stardoc: http://skydoc.bazel.build -->
+
+<a name="#dotnet_wrapper"></a>
 
-TODO: generate with stardoc (issue [#17](https://github.com/Brightspace/rules_csharp/issues/17)).
+## dotnet_wrapper
```
j3parker: Could we hide this from the output?
I get that `dotnet_wrapper` needs to be public for Reasons(TM) but I don't think it's an API we want to document.

j3parker:
Cool!

docs/APIReference.md
jrbeverly: The rule `stardoc` offers a `symbol_names` attribute where you can list all of the symbols you want to have documented.

```bazel
stardoc(
    name = "my-docs",
    input = "doc_hub.bzl",
    out = "docs.md",
    symbol_names = ["foo_rule", "bar_rule", "baz_rule"],
)
```

I think we would want to restrict the list using that.

csharp/private/actions/BUILD
jrbeverly: The usages of `bzl_library` are for aggregating rules files together, so that actions can be applied on them (like tests or documentation).  Setting the default visibility is a quick way of allowing this to work (without it we cannot find the '.bzl' files) for docs+tests. You can see a similar usage in [rules_python](https://github.com/bazelbuild/rules_python/blob/master/python/BUILD#L27)

I think we may be able to limit the visibility to `visibility = ["//docs:__pkg__"]` and use `exports_files` for exposing the rules files, but I'll need to delve into that a bit more first.

jrbeverly: After looking at how `bazel_tools` handles this, I think I may be able to use a combination of `filegroups` and restricted visibility to remove the need for the default visibility of public.

jrbeverly: This was successful at limiting visibility permissions, as well as cleaning up usages in all of the files.

csharp/defs.for_docs.bzl
```diff
@@ -0,0 +1,47 @@
...
+csharp_library = _csharp_library
+csharp_library_set = _csharp_library_set
+csharp_nunit_test = _csharp_nunit_test
+csharp_register_toolchains = _csharp_register_toolchains
+csharp_repositories = _csharp_repositories
+import_multiframework_library = _import_multiframework_library
+import_library = _import_library
+import_nuget_package = _import_nuget_package
+nuget_package = _nuget_package
+setup_basic_nuget_package = _setup_basic_nuget_package
```
j3parker:
```suggestion
```

This one is another internal detail that I'd be happy to not have documented (other than in the code). It's public for Reasons(TM) though.

PR #112, opened 2019-12-11
Aggregate the bazel rules files for use in tooling

jrbeverly:
Aggregate all of the bazel sources together as using the model of `bzl_srcs`. This allows us to configure docs or tools on the bazel rule themselves. The top level is not accessible to any packages (e.g. `//docs`), as that will be handled in #110.

This PR aggregates all of the bazel sources to the top level using filegroups and the `bzl_srcs` model. This allows us to run tooling against our bazel rules files.

Relates to #17

### DIFF
--- a/csharp/BUILD
+++ b/csharp/BUILD
@@ -1,0 +1,6 @@
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["*.bzl"]) + [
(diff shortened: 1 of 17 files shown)


jrbeverly:
Opened this with the WIP text before I had a chance to fill in the right stuff 🤷‍♂

Issue #101, opened 2019-11-11
Should we pin to a version of Bazel in CI?


j3parker:
Currently we pin to a version of Bazel in CI.

Most projects in the bazelbuild org don't do this (a couple do, though). Some pin for specific configurations (e.g. to check support for a "minimum Bazel version").

The benefit of pinning is that we can opt-in to breakage in a particular PR. The downside is that we want to move quickly. A mitigating factor is that post-1.0 the intent is for Bazel to be reasonable about the pace of breaking changes.

jrbeverly:
I'd prefer it if we didn't pin the version of bazel in CI at the moment. When the rules have reached (or close to) D2L generally usable, I'd be more in favour of pinning them.

If we pin the version now, we will need to wait for a version bump to identify any breaking changes (or deprecation notices). How frequently that occurs requires dependabot, or someone to be on top of updating bazel in CI.

Not pinning isn't without risks (such as a breaking change failing every pull request), but I'd be more confident with the rules if we were forced to always be at the latest version.

j3parker:
Closing in favour of #155 which will allow us to mitigate the downsides of pinning.

PR #113, opened 2019-12-11
Run buildifier in CI to keep formatting

jrbeverly:
Requires that all WORKSPACE, BUILD, and .bzl files meet lint + formatting rules of buildifier. This is pinned to the latest version.

You can see the documentation for this [here](https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#running-buildifier-on-ci).

Relates to #101, #111 

Additionally I performed a format of the YAML file, but there isn't any format guideline set in the code like `.clang-format`.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -3,6 +3,7 @@ bazel: 1.2.1
 
 buildifier:
   version: latest
+  warnings: disable=module-docstring,disable=function-docstring,disable=function-docstring-args
 
 tasks:
   examples-windows:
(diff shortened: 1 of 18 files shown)


j3parker:
Feel free to add an `.editorconfig`, maybe

jrbeverly:
I've taken a look at the warnings that are being thrown at the moment, and I'll be trying to suppress those warnings for the time being. I primarily want buildifier running in CI to prevent issues with syntax formatting.

To resolve all of the linting errors, we will need to be looking at documenting functions, files and arguments. Additionally some renaming will need to take place in `providers.bzl` to follow the constant and provider naming conventions.

PR #117, opened 2019-12-19
Fixes for update-docs.sh

j3parker:


update-docs.sh
```diff
@@ -3,4 +3,6 @@
 set -euo pipefail
 
 bazel build //docs/...
-cp bazel-bin/docs/*.md docs/
\ No newline at end of file
+
+# Copy the file, ignoring the permissions inside bazel-bin
+cat bazel-bin/docs/APIReference.md > docs/APIReference.md
```
j3parker: Inside bazel-bin files are executable and not writable which is annoying.

What this does exactly depends on your umask, but git only tracks the most basic info about file permissions so in general this should be good.

WORKSPACE
```diff
@@ -23,7 +23,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 git_repository(
     name = "io_bazel_stardoc",
     remote = "https://github.com/bazelbuild/stardoc.git",
-    tag = "0.4.0",
+    commit = "4378e9b6bb2831de7143580594782f538f461180",
+    shallow_since = "1570829166 -0400",
```
j3parker: Tags are mutable, so Bazel complains/tells you to do this if you reference external things by tag.
(`shallow_since` is an optimization; I just copied this from the DEBUG output).

j3parker:
I'm going to re-open from a fork since our CI is set up to assume that.

PR #120, opened 2019-12-19
Add runtimeconfig.json for .NET Core executables

j3parker:
This is half of #71. The other half is invoking the exe with `dotnet` rather than directly (I'm going to use the wrapper to do this, but in a follow-up PR).

### DIFF
--- a/examples/basic/BUILD
+++ b/examples/basic/BUILD
@@ -9,7 +9,7 @@ load(
 csharp_binary(
     name = "hello",
     srcs = ["hello.cs"],
-    target_frameworks = ["net472"],
+    target_frameworks = ["netcoreapp3.0"],
     deps = [":lib"],
 )
 
@@ -20,12 +20,12 @@ csharp_binary(
 csharp_nunit_test(
     name = "lib_test",
     srcs = ["libtest.cs"],
-    target_frameworks = ["net472"],
+    target_frameworks = ["netcoreapp3.0"],
     deps = [":lib"],
 )
 
 csharp_library(
     name = "lib",
     srcs = ["lib.cs"],
-    target_frameworks = ["net472"],
+    target_frameworks = ["netcoreapp3.0"],
 )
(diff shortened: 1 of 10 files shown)


examples/basic/BUILD
```diff
@@ -27,5 +27,5 @@ csharp_nunit_test(
 csharp_library(
     name = "lib",
     srcs = ["lib.cs"],
-    target_frameworks = ["net472"],
+    target_frameworks = ["netcoreapp3.0"],
```
j3parker: This only works nicely if the TFMs match: if it were .NET standard the DLL would be in another folder.

I'm now thinking that for .NET Core, rather than wrapping with a customized assembly loader maybe we should just add more probing paths to the runtimeconfig.json. We'll still need a wrapper for .NET Framework I think (see #9 for more details).

So my goal is to change this library to `netstandardx.y` in the future.

j3parker:
Still not 100% working:

```
$ bazel run basic:hello
INFO: Analyzed target //basic:hello (21 packages loaded, 2285 targets configured).
INFO: Found 1 target...
Target //basic:hello up-to-date:
  bazel-bin/basic/bazelout/netcoreapp3.0/hello.exe
  bazel-bin/basic/bazelout/netcoreapp3.0/hello.ref.exe
  bazel-bin/basic/bazelout/netcoreapp3.0/hello.pdb
INFO: Elapsed time: 0.773s, Critical Path: 0.01s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
/bin/bash: /private/var/tmp/_bazel_jacob/5ce6d685aeab8cd41f37e1f62e684c76/execroot/__main__/bazel-out/darwin-fastbuild/bin/basic/bazelout/netcoreapp3.0/hello.exe: cannot execute binary file
```

But now we're half-way there:

```
$ dotnet /private/var/tmp/_bazel_jacob/5ce6d685aeab8cd41f37e1f62e684c76/execroot/__main__/bazel-out/darwin-fastbuild/bin/basic/bazelout/netcoreapp3.0/hello.exe
Hello, world!
Some numbers for you:
        1, 1, 3, 5, 13, 21, 55, 89, 233, 377
```

PR #122, opened 2020-01-06
Upgrade to Bazel 2.0.0 in CI

jrbeverly:
Upgrade bazel to the latest version in CI.

The full release announcement can be found in the [Blog entry on Bazel 2.0](https://blog.bazel.build/2019/12/19/bazel-2.0.html), and the release notes [here](https://github.com/bazelbuild/bazel/releases/tag/2.0.0).

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,5 +1,5 @@
 ---
-bazel: 1.2.1
+bazel: 2.0.0
 
 buildifier:
   version: latest


jrbeverly:
@j3parker Seems that the check paths for build-kite have changed, so I have updated the branch rules to use these new checks (the ones above), instead of the old `/pr/` ones.

omsmith:
I would estimated that the check paths haven't changed. The ones without `/pr/` are just used when it's in-repo. I would suggest keeping the `/pr/` ones as required, and start doing your work from a fork (so it's equivalent to any third-party contributor).

jrbeverly:
Hmm, this PR was a one-off using the web 'edit' (all my other work is from my fork), but is this the intended behaviour of CI? 

If it is the intended behaviour, we should probably make a note when doing some improvements to the documentation to document this (maybe in CONTRIBUTING.md)

j3parker:
It's not desired behaviour but it is how it works :( I've messed this up quite often myself.

You can click the edit from your fork, if you remember.

Rather than documenting it we could just rely on branch protection rules. I've set every branch to require a PR (rather than just the master) which I think will solve this.

PR #123, opened 2020-01-07
Replace docs API with subset of rules

jrbeverly:
This PR removes the custom API for stardoc, instead using `symbol_names` to specify a subset of rules that we want documentation for.

`defs.for_docs.bzl` was a workaround to resolve issues with either the rules or the stardoc implementation. The blockers for using `symbol_names` have now been resolved, and this can be simplified to just a set of rule names.


jrbeverly:
I have switched the branch protections back to using the PR checks 🐉

PR #131, opened 2020-03-07
Generalize the dotnet wrapper a bit

j3parker:
`csharp_binary` needs to become a rule which emits a `cc_binary` that
invokes `dotnet <foo>`, where `<foo>` is the actual dll that we
compile, to solve #71. Eventually `<foo>` will become a wrapper C# exe
that tweaks assembly loading to solve #9. So all of this requires the
wrapper to provide a default `argv[1]`.

For stuff like `dotnet vstest` (#51) we will have `argv[1]` be `vstest`, so we
also need support for baking in `argv[2]`.

### DIFF
--- a/csharp/private/wrappers/dotnet.cc
+++ b/csharp/private/wrappers/dotnet.cc
@@ -1,3 +1,4 @@
+#include <cstring>
 #include <iostream>
 #include <sstream>
 #include <string>
@@ -21,17 +22,29 @@ std::string evprintf(std::string name, std::string path) {
   return ss.str();
 }
 
+// Bazel template strings. Optional args may == "".
+// kDotnetExe: path to dotnet.exe
+// kArgv1: (optional) first arg to dotnet.exe
+// kArgv2: (optional) second arg to dotnet.exe
+constexpr const char* kDotnetExe = "{DotnetExe}";
+constexpr const char* kArgv1 = "{Argv1}";
+constexpr const char* kArgv2 = "{Argv2}";
+
+static_assert(
+  !(kArgv1[0] == '\0' && kArgv2[0] != '\0'),
+  "kArgv2 can only  be specified if kArgv1 is too"
+);
+
 int main(int argc, char** argv) {
   std::string error;
 
   auto runfiles = Runfiles::Create(argv[0], &error);
-
   if (runfiles == nullptr) {
     std::cerr << "Couldn't load runfiles: " << error << std::endl;
     return 101;
   }
 
-  auto dotnet = runfiles->Rlocation("{DotnetExe}");
+  auto dotnet = runfiles->Rlocation(kDotnetExe);
   if (dotnet.empty()) {
     std::cerr << "Couldn't find the .NET runtime" << std::endl;
     return 404;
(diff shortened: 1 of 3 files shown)


j3parker:
Er, had this sitting around on a branch but I forgot to tidy some stuff up.

j3parker:
Ok.

j3parker:
Improvements for future PRs:

- Move template params out to a header to make it obvious where they are
- Remove the ifdefs in favour of Starlark [`select()`](https://docs.bazel.build/versions/master/be/functions.html#select) to chose the write helper method impl
- Switch to [`CreateProcess`](https://docs.microsoft.com/en-us/windows/win32/procthread/creating-processes) from spawnv on Windows.

csharp/private/wrappers/dotnet.cc
```diff
@@ -50,14 +63,34 @@ int main(int argc, char** argv) {
...
+
+  auto dotnet_argv = new char*[extra_count + argc + 1];
+
   // dotnet wants this to either be dotnet or dotnet.exe but doesn't have a
   // preference otherwise.
-  auto dotnet_argv = new char*[argc + 1];
-  dotnet_argv[0] = (char*)"dotnet";
+  dotnet_argv[0] = const_cast<char*>("dotnet");
+
+  // Make sure to hold a reference to these std::string.
```
omsmith: Because of the pointers from c_str?

csharp/private/wrappers/dotnet.cc
```diff
@@ -50,14 +63,34 @@ int main(int argc, char** argv) {
...
+  }
+
+  if (std::strlen(kArgv2) != 0) {
+    dotnet_argv[2] = const_cast<char*>(argv2.c_str());
+  }
+
+  // Copy the rest of our arguments in
   for (int i = 1; i < argc; i++) {
-    dotnet_argv[i] = argv[i];
+    dotnet_argv[i + extra_count] = argv[i];
```
omsmith: nit: prefer extra_count stayed consistently at the front

j3parker: Ah yeah good catch.

csharp/private/wrappers/dotnet.cc
j3parker: `Rlocation` returns an `std::string` and yeah, the output of `c_str()` is coupled to that lifetime.

csharp/private/wrappers/dotnet.cc
```diff
@@ -50,14 +63,34 @@ int main(int argc, char** argv) {
...
+  auto dotnet_argv = new char*[extra_count + argc + 1];
+
   // dotnet wants this to either be dotnet or dotnet.exe but doesn't have a
   // preference otherwise.
-  auto dotnet_argv = new char*[argc + 1];
-  dotnet_argv[0] = (char*)"dotnet";
+  dotnet_argv[0] = const_cast<char*>("dotnet");
+
+  // Make sure to hold a reference to these std::string.
+  auto argv1 = runfiles->Rlocation(kArgv1);
```
omsmith: https://github.com/bazelbuild/bazel/blob/46b285a8044d8c2dcf292f40b6b74bd99d66b430/tools/python/runfiles/runfiles.py#L138

is `""` `not`y?

j3parker: https://github.com/bazelbuild/bazel/blob/46b285a8044d8c2dcf292f40b6b74bd99d66b430/tools/cpp/runfiles/runfiles_src.cc#L170

PR #138, opened 2020-05-20
Fix new buildifer lints

j3parker:


j3parker:
(feel free to merge when you get to this, I'll be AFK)

PR #149, opened 2020-11-16
Correct .NET 5 runtime framework version

jimevans:
The runtime framework version for .NET 5 is not the same as the SDK
version. The SDK version is 5.0.100; the runtime framework version is
5.0.0. This commit addresses that discrepancy.


j3parker (commented):
😊

j3parker (approved):

Issue #140, opened 2020-10-27
Check for new .NET Core versions in a cronjob


j3parker:
Add a GitHub workflow that runs nightly to check [Microsoft.NETCore.App.Ref](https://www.nuget.org/packages/Microsoft.NETCore.App.Ref) for new releases. When a new release is found, open an issue. E.g. "Add support for netcoreapp5.0".

This job would need to look at which releases we do support by grepping one of the files or actually running some starlark.

What to do about preview releases? We could ignore them and just get an issue when a new release is GA. It might be nice though if we added support for new versions early via the preview packages (e.g. we could use `5.0.0-preview1234` for `5.0.0`.) If we do this we should have a second cronjob to see if there is a newer preview build/non-preview build available and submit a PR to update the URL/sha256.

omsmith:
Something the keep in mind:

> Warning: To prevent unnecessary workflow runs, scheduled workflows may be disabled automatically. When a public repository is forked, scheduled workflows are disabled by default. **In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days.**

Will probably require some manual action to keep it going (click a button in an e-mail) - unless the updates are actually happening often enough and thus we act on them and it stays active :)

j3parker:
Neat!
If we had a few more dependabot-like things, e.g. updating the Bazel version maybe we'd be OK.
Definitely something to keep in mind...

PR #163, opened 2021-04-05
update sdk to 5.0.104

omsmith:


omsmith:
There's also `5.0.201`, but I didn't look into differences. Just updating security patch here. I think we just always want latest though?

j3parker:
> I think we just always want latest though?

I think so.
Related: https://github.com/Brightspace/rules_csharp/issues/140

j3parker (dismissed):

